PR #40, opened 2020-03-09
Add tracing API

somewhatabstract:
# Summary

This adds an API for tracing events.

This is based off the `profile` API that `react-render-server` uses. I have worked to simplify the API based off lessons learned from using that API. This includes renaming it to `trace` so that it fits the nomenclature of StackDriver Trace, with which it integrates.

Usage:

`trace("The thing I want to trace")`
`trace("The thing I want to trace", request)`

The function returns an `ITraceSession` interface that provides the `name` of the trace session as provided in the `trace` call, and an `end` function for ending the trace session.

## Issues

This is a part of WEB-1858.

## Review Guidelines

Nothing is using this API yet. The intention will be to use this in a few different ways. For some examples, see the react-render-server code where we profile things like the render request, and the download of JS files. In the `render-gateway` we will be providing a wrapper of this API to the page rendering implementation of webapp so that we can also trace additional internal elements of our webapp rendering, such as data requests or component renders. This will allow us to not only see detailed logs of these events, but also view them in the StackDriver Trace user interface, enabling us to dig more deeply into SSR performance.

Note that StackDriver Trace does not like hundreds of trace elements, so we may want to consider a way to "tag" traces and then enable/disable them based on configuration. That way folks can annotate their code and we can enable tracing as we wish, like a "high level" trace versus a "render" trace. We could also consider only sending certain traced events to StackDriver, but still logging them all. This work has not been done yet as I think we want to see how this gets used before we start customizing to that level.


flow-typed/custom/@google-cloud/trace-agent.js
```diff
@@ -37,12 +37,15 @@ declare function traceagent$start(
...
+declare function traceagent$get(): traceagent$Tracer;
+
 declare module "@google-cloud/trace-agent" {
     declare export type Config = traceagent$Config;
     declare export type SpanOptions = traceagent$SpanOptions;
     declare export type Span = traceagent$Span;
     declare export type Tracer = traceagent$Tracer;
     declare module.exports: {
         start: typeof traceagent$start,
+        get: typeof traceagent$get,
```
somewhatabstract: This is part of the API, I just hadn't bothered adding it to the flow types before because we weren't using it.

src/ka-shared/trace.js
```diff
@@ -0,0 +1,33 @@
...
+ * fulfilled. This is used to determine if a request-scoped logger can be used.
+ * @returns {ITraceSession} The new trace session that was created and is to be
+ * used to end the session.
+ */
+export const trace = <TReq: RequestWithLog<$Request>>(
+    name: string,
+    request?: TReq,
+): ITraceSession => {
+    const logger = getLogger(request);
+    const tracer = traceAgent.get();
```
somewhatabstract: `traceAgent.get` is guaranteed to return a tracer, even if tracing is disabled. If tracing is disabled, it just won't actually lead to anything getting traced.

src/ka-shared/types.js
```diff
@@ -1,10 +1,38 @@
...
+ * A collection of secrets keyed by their names.
+ */
 export type Secrets = {[string]: SecretString, ...};
+
+/**
+ * Configuration of secrets lookup.
+ *
+ * One variation provides a Google Cloud KMS configuration; the other provides
+ * means to map secrets dynamically.
+ */
```
somewhatabstract: Ooops, this should be in my other PR on types stuff. I'll move it.

src/shared/__tests__/__snapshots__/trace.test.js.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`#trace should throw if the name is empty/falsy 1`] = `"Must provide a name for the trace session."`;
+
+exports[`#trace should throw if the name is empty/falsy 2`] = `"Must provide a name for the trace session."`;
+
+exports[`#trace should throw if the name is empty/falsy 3`] = `"Must provide a name for the trace session."`;
```
somewhatabstract: Can't do inline snapshots when using `it.each`

.prettierignore
```diff
@@ -2,4 +2,5 @@
 *.md
 docs
 *.css
-*.yml
\ No newline at end of file
+*.yml
+dist
```
somewhatabstract: Added `dist` to ignore list since we don't want prettier churning our built code.

src/shared/trace.js
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Next, if we were given a tracer, we start a trace section for this so
+     * trace session so that it will appear in Stackdriver Trace.
+     *
+     * We annotate the span with "TRACE:" so that it is clear in the trace
+     * which spans were created by this API and which were inserted by other
+     * means.
+     */
+    const span = tracer?.createChildSpan({name: `TRACE: ${name}`});
```
somewhatabstract: RRS used `PROFILE(start):`. I use `TRACE:` because it is much shorter and easier to scan for the real data of the trace name.

src/shared/trace.js
```diff
@@ -0,0 +1,91 @@
...
+    const end = (info?: TraceSessionInfo): void => {
+        /**
+         * Let's mark our profile as done.
+         *
+         * We include the session info object, but make sure to set the level
+         * and message ourselves.
+         */
+        profiler.done({
+            ...info,
+            message: `TRACED: ${name}`,
```
somewhatabstract: I used `PROFILE(end):` in the RRS. Here I settled on `TRACED:` as the tense implies it ended and it is much shorter. Should make logs and traces much easier to read.

src/shared/trace.js
```diff
@@ -0,0 +1,91 @@
...
+            level: info?.level || "debug",
+        });
+
+        /**
+         * If we started a tracer span, let's end it.
+         *
+         * We disable this lint rule as the linter does not appear to
+         * understand the optional chaining.
+         */
+        // eslint-disable-next-line flowtype/no-unused-expressions
```
somewhatabstract: I need to go see if there's an issue raised on their project about this.

src/shared/types.js
```diff
@@ -4,18 +4,57 @@ import type {
...
+};
+
+/**
+ * A trace session that has been started.
+ */
+export interface ITraceSession {
+    /**
+     * The name of the session as provided when it was started.
+     */
+    get name(): string;
```
somewhatabstract: The RRS version of a session did not include this, but I realised it may be useful to have, especially if we make this API available to more things and we want to include additional state info like whether the trace is ended yet, and if it is ended, how long it took.  For now, this should help with debugging at the very least.

src/shared/trace.js
```diff
@@ -0,0 +1,91 @@
...
+     */
+    const end = (info?: TraceSessionInfo): void => {
+        /**
+         * Let's mark our profile as done.
+         *
+         * We include the session info object, but make sure to set the level
+         * and message ourselves.
+         */
+        profiler.done({
+            ...info,
```
somewhatabstract: We pass the info here. This will attach it as metadata alongside the `durationMs` metadata that winston adds. This enables us to keep the logged message simply the trace event name, and have all the data embedded in the metadata.

PR #42, opened 2020-03-10
Improve trace API a little

somewhatabstract:
# Summary

This is a little tweak on top of #40.

In this PR, I have added the ability for the `trace` call to be invoked with a request or a log, or neither. 

## Issues

Relates to WEB-1858

## Review Guidelines

This helps with upcoming usage - I'm reworking my local changes into smaller PRs, so this has little context right now.


src/ka-shared/trace.js
```diff
@@ -3,31 +3,63 @@ import * as traceAgent from "@google-cloud/trace-agent";
...
+     * Note that if startTraceAgent was never called, this will still log but the
+     * StackDriver trace span creation will not actually happen.
+     *
+     * @param {string} name The name of the event being traced.
+     * @param {Logger} logger The logger to be used for the trace.
+     * @returns {ITraceSession} The new trace session that was created and is to be
+     * used to end the session.
+     */
+    (name: string, logger: Logger): ITraceSession;
+}
```
kevinbarabash: Cool!  I didn't realize you could have multiple call signatures on a callable.  I'm glad that Flow supports this.

somewhatabstract: We already do this in webapp too. Checkout the translation stuff I did for dev time i18n. :)

PR #43, opened 2020-03-10
Add dependencies and types for fetching work

somewhatabstract:
# Summary

This PR adds some new dependencies and some types that are to support the fetching implementation we will use for things like fetching JavaScript.

## Issues

This is a part of WEB-1858.

## Review Guidelines

The PRs stacked on this will show how these dependencies and types get used.


package.json
```diff
@@ -70,9 +70,12 @@
     "@google-cloud/logging-winston": "^3.0.6",
     "@google-cloud/profiler": "^3.0.1",
     "@google-cloud/trace-agent": "^4.2.5",
+    "agentkeepalive": "^4.1.0",
     "express": "^4.17.1",
     "express-async-handler": "^1.1.4",
     "express-winston": "^4.0.3",
+    "superagent": "^5.2.2",
+    "superagent-cache-plugin": "^3.0.0-beta.1",
```
jeresig: FWIW, I had a lot of trouble using the beta of this plugin, I really this we should try to stick to the 2.x branch until the final release is ready. The caching dynamics are totally changed - it's now based on headers/the response from the server and setting a custom cache time has no effect.

src/types.js
```diff
@@ -31,6 +31,110 @@ export type AuthenticationOptions = {
...
+     * The superagent-cache-plugin-compatible cache that will be used as the
+     * cache implementation.
+     */
+    provider: mixed,
+
+    /**
+     * A callback to calculate when the cached response for a given URL should
+     * expire. If this method is omitted, the cache provider's default
+     * expiration will be used. The result is given to superagent-cache-plugin
+     * and works according to its documentation.
```
jeresig: This is no longer true if we use v3

package.json
somewhatabstract: Oh, I never noticed it had installed the wrong one! Great catch, @jeresig

src/types.js
somewhatabstract: Yeah, I know. I didn't realise it had installed the beta.

PR #45, opened 2020-03-10
Implement extractErrorString

somewhatabstract:
# Summary

This implements the `extractErrorString` function which is a direct copy of the [`extractErrorInfo` method in react-render-server](https://github.com/Khan/react-render-server/blob/5709c900ef9a7b3fab36e354884f9b1ba98adaff/src/logging.js#L86-L104).

The difference here is that this is fully tested and has slightly stronger flow typing.

## Issues

Part of WEB-1848.


src/shared/extract-error-string.js
```diff
@@ -0,0 +1,32 @@
...
+        return error;
+    }
+
+    if (error.response && error.response.error) {
+        return `${error.response.error}: ${error.stack || ""}`;
+    }
+
+    if (error.error && error !== error.error) {
+        return extractErrorString(error.error);
+    }
```
kevinbarabash: I didn't realize errors could be recursive in nature.

src/shared/__tests__/extract-error-string.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+        // Assert
+        expect(result).toBe("THIS IS THE STACK!");
+    });
+
+    it("should return the result of toString if none of the other cases match", () => {
+        // Arrange
+        const error = {
+            this: "IS NOT LIKE THE OTHERS",
+            toString: () => "AND SO THIS IS THE ERROR!",
```
kevinbarabash: Using `this` as a key here isn't necessary is it?  It's a joke right?

somewhatabstract: Yeah, it's just a joke :grin:

src/shared/extract-error-string.js
somewhatabstract: Yeah, this was an exciting discovery as I tried to get RRS to output useful error info.

src/shared/__tests__/extract-error-string.test.js
kevinbarabash: It was funny, but seeing `this` as a key got my spidey sense tingling.  Please don't take the fact that I asked if it was a joke as it being unfunny. 😅

somewhatabstract: Can you imagine typing `this.this`? Imagine reviewing that code. lol

PR #46, opened 2020-03-10
Implement shouldRetry callback creation

somewhatabstract:
# Summary

Superagent supports retries. Rather than do them ourselves as we do in react-render-server, we are going to use its retry implementation. As part of it, it allows for a callback that is invoked before each retry (up to the max retries specified) so that we can both log the request failure, and if appropriate, prevent further retries.

We provide this as an overridable thing in the gateway options so that we can experiment from webapp without changing this code, should the need arise, which it probably will based off experiences with the RRS.

## Issues

Part of WEB-1848.

## Review Guidelines

With this last bit of infrastructure in place, the next diffs will start pulling things together.


src/make-should-retry.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+import type {Response as SuperAgentResponse, CallbackHandler} from "superagent";
```
somewhatabstract: These are all `any` because we don't have superagent flow types. I hope to rectify that at some point. For now, this at least helps annotate the code more clearly.

PR #47, opened 2020-03-10
Implement makeUnbufferedNoCacheRequest

somewhatabstract:
# Summary

This is the base request that we always perform whether we're caching or not. It sets up the agent, the URL being requested, the User-Agent header (super helpful when querying CDN logs, for example), timeout, and retry strategy.

The request made by this call will then be wrapped depending on the usage as we shall see in the next PRs. This is based off the requests we perform in `react-render-server`, but with retries now handled by superagent rather than our own retry logic.

## Issues

Part of WEB-1858


src/make-unbuffered-no-cache-request.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+export const makeUnbufferedNoCacheRequest = (
+    options: RenderGatewayOptions,
+    url: string,
+    logger: Logger,
+): SuperAgentRequest => {
+    const {name: gatewayName, requests: requestOptions} = options;
+
+    // Get an agent.
+    const agent = makeAgent(requestOptions?.keepAlive);
```
somewhatabstract: The RRS makes an agent once and then shares it across all requests. We'll likely need to do that too. I'll work on that once the pieces are in place.

dist/make-unbuffered-no-cache-request.js.map
```diff
@@ -0,0 +1 @@
...       * We give it a callback so we can log the retry and, if we so choose\n             * in the future, decide whether we should allow any more. This would\n             * allow us to short circuit the retry count (the max retries still\n             * takes precedence over our callback response, so we can't retry\n             * forever).\n             */\n            .retry(\n                requestOptions?.retries || 2,\n                makeShouldRetry(logger, requestOptions?.shouldRetry),\n            )\n            /**\n             * We add a user agent header so that we can easily identify our\n             * requests in logs.\n             *\n             * The header has a form like:\n             *     gateway-name (GAE_VERSION_STRING_HERE)\n             */\n            .set(\n                \"User-Agent\",\n                `${gatewayName} (${process.env.GAE_VERSION || \"UNKNOWN\"})`,\n            )\n            /**\n             * Our default timeout is 1 minute, but we allow for it to be\n             * overridden by gateway options.\n             */\n            .timeout(requestOptions?.timeout || 60000)\n    );\n};\n"],"file":"make-unbuffered-no-cache-request.js"}
```
kevinbarabash: Is this for debugging purposes in production?

src/make-unbuffered-no-cache-request.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+import superagent from "superagent";
+import type {SuperAgentRequest} from "superagent";
+import {makeAgent} from "./make-agent.js";
+import type {RenderGatewayOptions} from "./types.js";
+import {makeShouldRetry} from "./make-should-retry.js";
+import type {Logger} from "./shared/index.js";
```
kevinbarabash: style nit: in other projects we separate type imports from non-type imports.

src/__tests__/make-unbuffered-no-cache-request.test.js
```diff
@@ -0,0 +1,319 @@
...
+    });
+
+    it("should set the User-Agent header with gateway name and UNKNOWN if GAE_VERSION not set", () => {
+        // Arrange
+        /**
+         * We have to delete the env var.
+         * If we just set it to `undefined` it actually gets set to the string
+         * "undefined", which is not what we want at all.
+         */
+        delete process.env.GAE_VERSION;
```
kevinbarabash: Thanks for adding a comment.

dist/make-unbuffered-no-cache-request.js.map
somewhatabstract: Production, development. The built code is what runs for webapp folks.

src/make-unbuffered-no-cache-request.js
somewhatabstract: This is the default sorting eslint that is in eslint-config.

kevinbarabash: I didn't realize that there were sorting rules in our shared eslint-config. 😅

PR #49, opened 2020-03-10
Implement makeRequest

somewhatabstract:
# Summary

Here we bring together all our request pieces to create the main request function.

With this in place, we can:

- implementing WEB-1856 to download the manifest
- implementing the rest of WEB-1858 to perform the rendering flow
- rework things to get the agent and cache plugin instances shared across requests
- consider adding `trace` calls to each request

## Issues

Part of WEB-1856 and WEB-1858


src/make-request.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {boolean} [buffer] Defaults to true. When true, the response body will
+ * be buffered, otherwise it will not.
+ * @returns {SuperAgentRequest} A superagent request for the URL.
+ */
+export const makeRequest = (
+    options: RenderGatewayOptions,
+    url: string,
+    logger: Logger,
+    buffer?: boolean = true,
+): SuperAgentRequest => {
```
kevinbarabash: To actually make the request we'd do the following right:
```
const req = makeRequest(...);
...
req.end();
```
Sorry for asking newbie questions but I'm unfamiliar with backend stuff. 😞

somewhatabstract: That depends on whether `buffer` is true or not.

If it is true, we can just `await` the request (that's the usage we'll be using). When it isn't true, we might need to call `end`. You can checkout the superagent docs on that though.

Our usage is likely to be with `buffer` as true since that's what we do in RRS, but I wanted to add the flexibilty here as we might use this API for more requests than RRS does and I wasn't sure that all cases would like being forced to `buffer`.

kevinbarabash: It seems like that would be some methods like `set` would need to be called in the same event loop cycle as when the request was made while others like `prune` or `buffer` might be callable later (though it's probably not a good idea to do so).

somewhatabstract: It's not quite like that. The easiest analogy though perhaps inaccurate its to testing where describe and it and before all execute to register the operations and then the test state machine runs to execute those operations in a defined order.

Calls to buffer or end are what kick that state machine off.

kevinbarabash: I didn't realize `buffer` was similar to `end`.  That makes much more sense than what I was thinking.

somewhatabstract: Yeah, I think the `superagent` docs call out that you need to use one or the other.

PR #51, opened 2020-03-12
Add flowtypes for superagent

somewhatabstract:
# Summary

Well, this was exciting.

`superagent` is a complex beast because it does all sorts of magic when generating its API (like adding methods to the exported interface). I started out with the Typescript types and tried to make them work for flow (they use namespaces, which are notoriously hard to migrate).

In working on this, I discovered that our agent setup for `react-render-server` is wrong. Calling `agent` before `get` does something entirely different than calling it **after** `get`. Adding these flow types highlighted that and allowed us to fix it. I've added some inline comments relating to the other changes.

Of course, the other flow shortcoming is things like plugins that modify the original API. To work around this, I just added the `supercache-cache-plugin` types to the same file. There aren't many and they have simple APIs (I based them off the code for the plugin).

## Issues

Relates to various, including WEB-1858 and WEB-1856.


flow-typed/custom/superagent.js
```diff
@@ -0,0 +1,307 @@
...
+    type: string;
+    unauthorized: boolean;
+    xhr: XMLHttpRequest;
+    redirects: string[];
+}
+
+declare interface superagent$Request extends Promise<superagent$Response> {
+    cookies: string;
+    method: string;
+    url: string;
```
somewhatabstract: These were on the SuperAgentRequest type in the typescript code, but they are clearly added to all requests and we want them here, so I moved them.

flow-typed/custom/superagent.js
```diff
@@ -0,0 +1,307 @@
...
+    forceUpdate(forceUpdate: ?boolean): superagent$Request;
+    pruneQuery(pruneQuery: Array<string>): superagent$Request;
+    pruneHeader(pruneHeader: Array<string>): superagent$Request;
+    prune(
+        prune: (
+            response: superagent$Response,
+            gutResponse: (superagent$Response) => any,
+        ) => any,
+    ): superagent$Request;
+    responseProp(responseProp: string): superagent$Request;
```
somewhatabstract: `superagent-cache-plugin` types here.

src/__tests__/make-unbuffered-no-cache-request.test.js
```diff
@@ -18,7 +18,7 @@ describe("#makeUnbufferedNoCacheRequest", () => {
     it("should create an agent for the request", () => {
         // Arrange
         const fakeSuperagent = {
-            get: jest.fn().mockReturnThis(),
+            agent: jest.fn().mockReturnThis(),
```
somewhatabstract: We call `get` first now, which means we have to mock it at the import level. And vice versa, `agent` is called from the `get` result. Therefore, in all these tests we have swapped them over.

src/__tests__/make-unbuffered-no-cache-request.test.js
```diff
@@ -30,7 +30,7 @@ describe("#makeUnbufferedNoCacheRequest", () => {
             },
         };
         const fakeLogger: any = {};
-        jest.spyOn(Superagent, "agent").mockReturnValue(fakeSuperagent);
+        jest.spyOn(Superagent, "get").mockReturnValue(fakeSuperagent);
```
somewhatabstract: Here's the matching swap of `get` and `agent`.

src/make-unbuffered-no-cache-request.js
```diff
@@ -29,8 +29,8 @@ export const makeUnbufferedNoCacheRequest = (
     // Build our main fetcher using the configured agent.
     return (
         superagent
-            .agent(agent)
             .get(url)
+            .agent(agent)
```
somewhatabstract: Yup, this was not doing what we thought it should be.

src/requests-from-cache.js
```diff
@@ -1,9 +1,6 @@
 // @flow
```
somewhatabstract: @kevinbarabash the inclusion of these flowtypes should help make all this much clearer on how these calls get combined with others.

flow-typed/custom/superagent.js
```diff
@@ -0,0 +1,307 @@
+import * as fs from "fs";
+import * as http from "http";
+import * as stream from "stream";
+import * as cookiejar from "cookiejar";
+
+declare type superagent$CallbackHandler = (
+    err: any,
+    res: superagent$Response,
+) => ?boolean;
```
somewhatabstract: I changed this. The typescript defs have this returning void, but I know from the `retry` method that it can return a maybe boolean in order to specify the `shouldRetry` callback.

src/make-unbuffered-no-cache-request.js
jeresig: Wow, very interesting!

somewhatabstract: When I was trying to get the flow errors and type definitions to reconcile, I noticed the difference. That led me to read the actual code and see that there is a difference. That helped me determine what the flow types should be and how to fix the errors.

What kind of API does this though? Where the ordering changes a method's meaning? Like, change the method name if that's the case. What the heck?

src/requests-from-cache.js
```diff
@@ -28,16 +25,16 @@ export const isFromCache = (response: SuperAgentResponse): boolean =>
  * The request will resolve with an additional property to indicate if it was
  * resolved from cache or not.
  *
- * @param {SuperAgentRequest} request The request to be modified.
+ * @param {Request} request The request to be modified.
  * @param {boolean} buffer When true, the response body will be buffered,
  * otherwise it will not.
- * @returns {SuperAgentRequest} A superagent request supporting caching for the
+ * @returns {Promise<Response>} A superagent request supporting caching for the
  * given URL.
```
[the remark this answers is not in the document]
somewhatabstract: Yeah, understandable. The comment was obviously from before the flow types were correct.

I wonder if we could word this better, but for now, I think it is OK. I'll think on it and perhaps make some changes if I could up with a better wording.

PR #52, opened 2020-03-12
Fix flowconfig

somewhatabstract:
# Summary

Turns out the `flow check` run was not working right. Some flow commands seem to be OK with the `<PROJECT_ROOT>` thing in the config file, but `flow check` is not and was not doing the right thing.


somewhatabstract:
> LGTM. I wonder if we'll have to make this change in other projects once we get this version of flow. 🤔

It's been a problem for a while. It is likely broken in the version being used:

https://github.com/facebook/flow/issues/2989

PR #53, opened 2020-03-12
Make agent and cache shareable across requests

somewhatabstract:
# Summary

This refactors our superagent request stuff so that we can share the same agent and cache instances across requests. To do so, I have added a `RequestOptions` type that can be constructed from the gateway's `RequestsOptions`, replacing the "maybe"-ness of things to concrete agent and cache instances, as well as defaults.

We will be able to create one of these options objects at the start and reuse it as-is, or create slightly modified versions as needed when performing requests.

This refactoring hopefully simplifies APIs and tests. It also puts all our request defaults into a single place.

## Issues

WEB-1858, WEB-1856


src/__tests__/create-request-options.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+        // Assert
+        expect(superagentCachePluginSpy).toHaveBeenCalledWith("FAKE_PROVIDER");
+    });
+
+    it("should include options, agent, and plugin in created options", () => {
+        // Arrange
+        const fakeRequestsOptions: RequestsOptions = {
+            retries: 10,
+            timeout: 1000000000, // dollars, muhahahahhahahahahaaa 🤔
```
jeresig: lol

PR #54, opened 2020-03-13
Implement in-flight request sharing, tracing, etc.

somewhatabstract:
# Summary

This adds the in-flight request tracking that RRS has, as well as tracing for requests. Together this mostly completes the `fetchPackage`-type behavior from RRS.

The final step will be actual usage where we combine this with the request handler to get an appropriate logger, as well as the `createRequestOptions` call to get the agent and cache, etc. 

## Issues

WEB-1858, WEB-1856


src/request.js
```diff
@@ -0,0 +1,91 @@
...
+import {isFromCache} from "./requests-from-cache.js";
+import {trace} from "./ka-shared/index.js";
+
+/**
+ * This tracks our inflight requests.
+ */
+const inFlightRequests: InFlightRequests = {};
+
+/**
+ * Abort any requests that are inflight and clears the inflight request queue.
```
somewhatabstract: NIT: Tense is wrong on "clears"

PR #56, opened 2020-03-13
Fix security audit vulnerabilities

somewhatabstract:
# Summary

There are a couple of moderate vulnerabilities in the code. Upgrading eslint fixes most. The last vulnerability needs a resolution directive in the package.json.

I have turned on automated vulnerability PRs for future issues, which should reduce overhead of managing this (I am sure dependabot would have dealt with the eslint ones too).

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,5 +73,8 @@
     "express-async-handler": "^1.1.4",
     "express-winston": "^4.0.3",
     "winston": "^3.2.1"
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,5 +73,8 @@
     "express-async-handler": "^1.1.4",
     "express-winston": "^4.0.3",
     "winston": "^3.2.1"
+  },
+  "resolutions": {
```
jeresig: Interesting!

PR #58, opened 2020-03-14
Encapsulate gateway info

somewhatabstract:
# Summary

I was not keen on how much the various GAE env vars were propagated around the code, nor the lack of easily getting the gateway name from the passed options.

This encapsulates the values we use into a single place, abstracting them away from the rest of the code. It also ensures that the options provided name is made available in cases where the env vars are not set.

We can use this to encapsulate other global, static info as we need (for example, we could use this to replace the `getRuntimeMode` stuff if we really wanted, or at least expose that off this call as well).


src/shared/get-gateway-info.js
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Get info about the running gateway.
+ *
+ * Encapsulates the retrieval of gateway information to abstract away things
+ * like GAE env vars.
+ *
+ * @returns {GatewayInfo} The information about the gateway.
+ */
+export const getGatewayInfo = (): GatewayInfo => ({
```
jeresig: Very nice, this makes it easier to customize (in the future), as well!

PR #60, opened 2020-03-18
Implement basic rendering flow

somewhatabstract:
# Summary

This PR solidifies how we're going to invoke the general render architecture. The way this works is that we are putting the burden onto the calling code (of course, we'll support it in this repo with the necessary pieces as appropriate).

So, instead of passing all those options in to control requests and the like, we just get a function that, given a request URL and the headers for it, will give us a result. The calling code is then responsible for things like:

- Environment creation (the VM/JSDOM/etc.)
- File downloads
- Caching

We will support this things with exports of common implementations for reuse, but this way the calling code can experiment with alternatives as needed.

The result of the render is a `body` representing the main result (i.e. our HTML) and a `status`, which depends on the status code. The `status` is of a type that allows us to encapsulate the various different status codes we support and their optional headers. There are then standard helpers for creating these valid codes, ensuring we get the right info.

Upcoming diffs will add the additional things such as:
- Proper header access tracking and Vary header creation
- Provision of trace API to the render callback for deep tracing into the render process
- Proper setting of the render path and request URL (currently, we're just passing the whole URL but in reality, we probably want to just pass the value of some preset query param)

## Issues

WEB-1858

## Review Guidelines

I have made each commit a distinct subset of the changes. So, rather than my previous work where each PR was a subset of the changes, now they are all in the same PR but each commit can be used to break it down as needed when reviewing. The breakdown is as follows:

### 1:Update VS Code settings for todo tree

Change to VS Code settings to support todo tree extension in this workspace, filtering out flow-typed and dist folders to reduce noise/increase signal.

### 2:Refactor extractErrorString in extractError

`extractErrorString` was a hangover from `react-render-server`. In writing this work, I realised that it would be more helpful to have a `SimplifiedError` type instead. So I refactored `extractErrorString` into the similar but not quite the same `extractError`.

### 3:Move to render callback architecture

We were getting options overload and in trying to work out how to get the `request` stuff shared between this core implementation and the calling code that would be using it, I realised that the core implementation likely doesn't need it at all. Instead, we want to ask the calling code to render and it should give us an answer.

This way, the eventual render implementation could be something we provide a common implementation for or some experimental stuff in a new webapp diff. This is a much more flexible and simpler architecture.

### 4:Add status codes

With the new architecture defined, we needed a way to encapsulate status codes so that, for example, a redirect code enforced having a redirect location header. It also enables us to enforce some rules on what status codes are permitted. For now, I have implemented 404, 200, and 301/302 codes. We should not need any others at the SSR level - if someone tries to SSR a route that requires permissions, I think we should 404 it since I don't think we can 401 as we won't be adding the appropriate headers that code demands (we could 403 it though that is perhaps incorrect since technically at the server, no authorization has yet occurred).

### 5:Implement render handler creation

Here we tie this together into our render handler. We replace the naïve render handler with a creation function that  makes a handler around the render callback. If we look at the `react-render-server`, this encapsulates all we need as everything else is the responsibility of the render callback implementation.

The request stats that were being logged by `react-render-server` can still be generated if we really need them, though I suspect that we do not; at least not in their current guise, so I have dropped that middleware for now with a view to wiring in what we need as we continue the `render-gateway` implementation.

### 7:Simplify render callback results

This rolls back some of the status code work to give more autonomy to the render implementation regarding what headers are included. The gateway doesn't yet use these headers. That will get added along with the Vary header work in an upcoming PR.

### 6:Update build artifacts and 8

This is just updates to `dist` to avoid the Github Action needing to work.

### Next up

To give some idea of where this heads next, we will be implementing some of the helper code that a calling gateway service can use to implement their render callback. So, in addition to the already written `request` API, which includes caching things, we can add render environment creation and setup, etc.

We will also add a proper header access tracker and Vary header creation, and tracing of the render request and passing that into the render callback itself.

All that said, the big news here is that "technically" this version of the render gateway could render a page if the render callback did something, as evidenced by the simple `run_local.js` variant. 🎉


somewhatabstract:
The small decrease in coverage is down to the header tracking function that is currently untested. I didn't write tests for that as it will soon be gone in favour of some fully tested code around a class that will track headers and produce a Vary header. Then we will just test that this gets used in this code and is passed appropriately.

src/status-codes.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Create a Redirect status.
+ *
+ * @param {string} targetURL The target URL of the redirect.
+ * @param {boolean} [isPermanent] If truthy, the status code 301 will be used to
+ * indicate a permanent move of the redirected URL; otherwise, the status code
+ * 302 will be used, indicating a temporary move. Defaults to false.
+ */
+export const redirect = (
```
jeresig: Hmm - I actually don't think this is how we want to implement this, I think this logic should be encapsulated inside the webapp rendering logic (namely that internally we'll know if we're redirecting, we'll return a status code of 301 or 302 and a correct Location header, but this system won't need to know that we "need to do a redirect" - it only needs to care about 3 things: status code, a list of headers, and the body text).

src/status-codes.js
```diff
@@ -0,0 +1,51 @@
...
+const NotFoundImpl: NotFoundStatus = {
+    get code(): 404 {
+        return 404;
+    },
+};
+
+/**
+ * Create a 404 NotFound status.
+ */
+export const notFound = (): NotFoundStatus => NotFoundImpl;
```
jeresig: In what scenario would the server know to generate a 404 error? I feel like that needs to be left up to the rendering logic to handle. It seems much more likely to me that we need to have a 500 status that we can return if there's an exception during rendering - there render-gateway isn't going to know if some URL does, or doesn't, exist, right?

src/handlers/make-render-handler.js
```diff
@@ -0,0 +1,79 @@
...
+         * Defer this bit to the render callback.
+         */
+        const {body, status} = await renderFn(renderURL, trackHeaderLookup);
+
+        /**
+         * If the status is a redirect, we need to set the redirect header/
+         */
+        if (status.code === 301 || status.code === 302) {
+            res.header("Location", status.targetURL);
+        }
```
jeresig: I mentioned this below but I think we can just remove this - and instead add in logic for handling a map of headers in general. There are a lot of headers that we're going to need to manage during rendering and it doesn't make sense to have them be a part of the render-gateway.

src/handlers/make-render-handler.js
```diff
@@ -0,0 +1,79 @@
...
+    res: Response,
+): Promise<void> {
+    const logger = getLogger(req);
+
+    // TODO(somewhatabstract): Actually track headers and build vary header.
+    //                         Encapsulate in other code to make it easily
+    //                         tested.
+    const trackHeaderLookup = (name: string): ?string => {
+        return req.header(name);
+    };
```
jeresig: Yeah, I like this model - I think this makes sense to me!

src/status-codes.js
somewhatabstract: At some point, the webapp code has to signal back to the gateway what code to return, including 404s where our routing didn't match the requested route. This will be part of the API, but yeah, we can certainly simplify how that will work, though I worry about allowing just anything - I wanted to build in some rigor so that it's easy to get it right.

That said, perhaps we can do that by providing a different set of helpers to build a render result that encourage supplying the correct headers.

src/handlers/make-render-handler.js
somewhatabstract: Are there a lot of headers to manage? What other headers get added to the response?

src/status-codes.js
jeresig: IMO, I'm fine deferring the question of rigor until much later - I think for now most of this logic is going to be managed in PageSettings anyway so we can just pass back the status code + headers + body and keep render-gateway nice and simple!

src/handlers/make-render-handler.js
jeresig: Ones that I'm aware of from `web/request/request_handler.py`: Location, X-Frame-Options, X-Content-Type-Options - but I'm fairly certain we set other headers in other handlers - doing a quick search I see a lot of "Content-Type" and "Content-Disposition" (but maybe those will go to the REST gateway) but there are also "Cache-Control", "Pragma", "Content-Security-Policy", and "Surrogate-Control". I also see a few Set-Cookie but I think we want to move most of those to the client-side. So, in short, there may be a few but it makes more sense for us to manage those in the rendering logic rather than try to push them into the render-gateway and handle every eventuality - easier to just blanket accept headers.

somewhatabstract: Yeah, I was planning on allowing more. The idea was to ensure that we didn't, say try to return a 301 without the header, but I think I can enforce that differently.

src/__tests__/index.test.js
```diff
@@ -10,4 +10,15 @@ describe("index.js", () => {
         // Assert
         expect(result).toContainAllKeys(["runServer"]);
     });
+
+    // it("should export status codes we expect", async () => {
```
jeresig: Commented out? Looks like it might be gone now.

somewhatabstract: Oh, good  catch. Thanks :)

PR #63, opened 2020-03-19
Distribute flow types

somewhatabstract:
# Summary

This adds a build step to output `.flow` files alongside our exported JS.

Since we export both `index.js` files and a couple of others from some packages, I have introduced a way to specify these additional export files from within the `index.js` so that we can document exports in a single place that is front and center to the maintained codebase, rather than buried in a hard-coded list in a build tool script. This also prevents us from just exporting a `.flow` file for every single source file.

The build tool reads this info from the `index.js` of each "package" and writes out the files accordingly.

## Issues

This is so that we can rely on types for WEB-1856 work inside the webapp repo.

## Details

To make this easier and also just make code easier to read/maintain, the gateway code is now in its own folder rather than at the root of `src`.

### Test Plan

To test this, I took the following steps.

1. Changed the import in the `run_local.js` file to use `dist` instead of `src`. Flow immediately complains about this because it cannot get the types.
2. Run the package.json command to build the code and types `yarn build`
3. Modify the `.flowconfig` to remove the `dist` folder from the `[Ignore]` section
4. Edit the `run_local.js` to say, pass the `port` as a string and see flow errors now occur


tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+ * Generates *.js.flow files in dist for each file that is an export for
+ * a given package. It means we have to ship source code but we're doing that
+ * anyway right now, so no worries.
+ *
+ * Besides index.js files, we also export some other specific files for
+ * general consumption. In order to make this scalable, but reasonably fast
+ * we assume that each index.js file has a header in it that identifies
+ * additional export files. That way we only need to look for index.js files
+ * and read those to work out the rest. This makes it at least a little
+ * scalable but also self-documenting.
```
kevinbarabash: Nice!

tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+ */
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+
+/**
+ * Some regex's we use to support our additional-files syntax.
+ */
+/**
+ * The first extracts the @additional-exports block, which looks like this:
+ *
```
kevinbarabash: style nit: these comment blocks could be joined.

tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+ *    "./fileb.js",
+ * ]
+ *
+ * Paths in that block are assumed to be relative to the directory of the file
+ * declaring them.
+ *
+ * The second regex grabs the filenames themselves from this block.
+ */
+const regexAdditionalFiles = /^(?:\s*.*@additional-exports\s?\[)((?:\s*.*"(?:[./\-\w]+)",\s*)*)(?:\s*.*\])$/im;
+const regexExtractFile = /^(?:\s*.*"([./\-\w]+)",\s*)$/gim;
```
kevinbarabash: I'm going to assume that these are right.  My regex-fu is so bad.  😭

tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Write out a .js.flow file in our disk for the file specified.
+ */
+const writeFlowFile = (
+    exportFromPath /*: string*/,
+    outputPath /*: string*/,
+) /*: Promise<void>*/ => {
+    const contents = ["// @flow", `export * from "${exportFromPath}";`].join(
+        "\n",
+    );
```
kevinbarabash: Will this work with a module that does a default export?

tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+     */
+    const packages = fs.readdirSync(distDir);
+    console.log(`Building flow files for ${packages.join(", ")}`);
+    for (const pkg of packages) {
+        const pkgSrcDir = path.join(srcDir, pkg);
+        const pkgDistDir = path.join(distDir, pkg);
+
+        const srcIndexJSPath = path.join(pkgSrcDir, "index.js");
+        const filesThatNeedTypes = await getExportsFrom(srcIndexJSPath);
+        for (const file of filesThatNeedTypes) {
```
kevinbarabash: This doesn't handle `index.js` files.  Are those handled manually at the moment?

tools/build-flow-types.js
```diff
@@ -0,0 +1,155 @@
...
+            const fileInSrc = path.normalize(path.join(pkgSrcDir, file));
+            const fileInDist = path.normalize(path.join(pkgDistDir, file));
+            if (!fs.existsSync(fileInSrc)) {
+                throw new Error(
+                    `Invalid file path in @additional-files block for "${path.relative(
+                        rootDir,
+                        srcIndexJSPath,
+                    )}": "${file}"`,
+                );
```
kevinbarabash: nice

tools/build-flow-types.js
somewhatabstract: I kept them separate so that hopefully, the intellisense doesn't include the bit that's only for documenting locally in the file

tools/build-flow-types.js
somewhatabstract: I copied it from Wonder Blocks. I sure hope so

tools/build-flow-types.js
somewhatabstract: It does. See the `getExportsFrom` method

kevinbarabash: :+1: I missed the `const files = [path.basename(filePath)];`.

tools/build-flow-types.js
somewhatabstract: I used regex101.com to create and test these. You could do the same to check them in a review.

kevinbarabash: That's a good point.  If there are particular case I was worried about I could double check that it works with those.

somewhatabstract:
> Could you add a "Test Plan" with a brief outlines what you ran to generate the types and how you verified that flow was picking them up as expected?

Sure.

PR #64, opened 2020-03-19
Export Requests API so that our service implementation can use it

somewhatabstract:
# Summary

As I was working on the manifest and file downloads, I realised that this really needs to be exported as a first class part of the package, rather than exporting from a specific file. So, I'm doing just that.

## Issues

WEB-1856


src/gateway/index.js
```diff
@@ -5,7 +5,9 @@
  */
 // eslint-disable-next-line import/no-unassigned-import
 import "../ka-shared/start-trace-agent.js";
+import * as RequestAPI from "./request.js";
 
-export type {RenderGatewayOptions} from "./types.js";
+export type {RenderGatewayOptions, RequestOptions} from "./types.js";
```
somewhatabstract: Should we export `AbortablePromise` too? Probably.

src/gateway/index.js
```diff
@@ -5,7 +5,9 @@
...
 
 export {runServer} from "./run-server.js";
+export const Requests = RequestAPI;
```
kevinbarabash: I like this kind of namespacing.

PR #65, opened 2020-03-20
Implement basic API

somewhatabstract:
# Summary

This puts the simple render API concept into place. Currently we just give two calls. A header accessor and a trace function. We'll likely want to add the logger too.

## Issues

WEB-1856


run_local.js
```diff
@@ -9,15 +9,19 @@
...
 async function main() {
     runServer({
         name: "DEV_LOCAL",
         port: 8080,
         // TODO(somewhatabstract): Implement something more complete for testing.
-        renderFn: () =>
+        renderFn: (
+            url /*: string*/,
+            renderAPI /*: RenderAPI*/,
+        ) /*: Promise<RenderResult>*/ =>
```
kevinbarabash: Why does this file use comments for flow annotations?

somewhatabstract: Because it's run directly by node. It's not built first.

PR #66, opened 2020-03-20
[WEB-1856] Render route using url param

somewhatabstract:
## Summary

This updates the routing to have a specific `/render` route with a
`url` param for specifying the URL being rendered.

## Issues

WEB-1856


src/gateway/handlers/make-render-handler.js
```diff
@@ -39,11 +39,15 @@ async function renderHandler(
     const traceFn = (name: string): ITraceSession => trace(name, req);
 
     /**
-     * TODO(somewhatabstract, WEB-1856): Currently passing the entire URL, but
-     * we want to be more specific here and define the render route better as
-     * we'll really want an absolute URL.
+     * The URL being rendered is given in a query param named, url.
      */
-    const renderURL = req.url;
+    const renderURL = req.query["url"];
```
jeresig: Can just be: `req.query.url`

somewhatabstract: I was following the express examples about this that seemed to favor the indexing approach.

PR #67, opened 2020-03-21
Export more request types/API and add logger to RenderAPI

somewhatabstract:
# Summary

As I'm implementing the service, I have noticed some missing pieces from the exports, so this adds those. It also adds the logger to the `RenderAPI`.

## Issues

WEB-1856


src/gateway/types.js
```diff
@@ -234,6 +234,11 @@ export type RenderAPI = {
      * Callback to start a trace session for tracing an operation.
      */
     +trace: TraceCallback,
+
+    /**
+     * A logger to use for logging during the render operation.
+     */
+    +logger: Logger,
```
kevinbarabash: While the whole type could be wrapped in `$ReadOnly<{ ... }>`, I can totally understand wanting to use `+` from a documentation point of view.

somewhatabstract: Yeah, I find `$ReadOnly` kinda cumbersome at the type level. I'd really love a `readonly` keyword like C# has.

PR #77, opened 2020-03-25
Connectivity tweakery

somewhatabstract:
## Summary

This makes a few modifications to get our connectivity working:

1. Removed `agentkeepalive` as it didn't appear to work. However, that may have been because of another bug I found much later in the webapp service side of things. That said, removing this makes things a lot simpler, and if we decide we want/need to try this again, it will be much easier to just use superagent's built in http/2 support, which should be equivalent, easier, and better all round.
2. Remove `RequestsOptions` and have agent be passed in via `RequestOptions`
3. Add a `getAgentForURL` method like `graphql-gateway` has but with some tweaks and export so that users of our Requests API can use it.
4. Add the 502/Load Balancer fix that `graphql-gateway` has.
5. Fix the error reporting in our `shouldRetry` callback. Turns out the response can be nully. I've updated the flowtypes for the `CallbackHandler` in superagent accordingly.
6. Fix the `shouldRetry` logging so that it only logs when there is an error.

## Issues

This is to support getting WEB-1856 working.

## Test Plan

Incorporating this build into phabD61770 and doing a manual deploy, then curling it to see that it works. (I'll be doing this bit tomorrow).

## Review Guidelines

Apologies, the commits go around the houses here while I experimented, but they tell a story that you should be able to follow if you want to review commit by commit.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "render-gateway",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
somewhatabstract: Seemed like it was worth a version bump

package.json
```diff
@@ -15,7 +15,8 @@
     "clean": "rm -rf dist",
     "test": "jest --colors --config jest.config.js src",
     "coverage": "jest --colors --config jest.config.js --coverage",
-    "lint": "eslint --report-unused-disable-directives --config .eslintrc.js \"{run_local.js,src,bin,__{tests,mocks}__}/**/*.js\"",
+    "lint:all": "yarn lint \"{run_local.js,src,bin,__{tests,mocks}__}/**/*.js\"",
+    "lint": "eslint --report-unused-disable-directives --config .eslintrc.js --ignore-path .eslintignore",
```
somewhatabstract: Our `lint-staged` was passing files to `lint` but we were still linting everything. So I moved the paths list to a `lint:all` command and left this without them (but added the ignore file).

run_local.js
```diff
@@ -21,7 +21,7 @@ async function main() {
             renderAPI /*: RenderAPI*/,
         ) /*: Promise<RenderResult>*/ =>
             Promise.resolve({
-                body: `THIS IS A RENDERED PAGE: ${url}`,
+                body: `You asked us to render ${url}`,
```
somewhatabstract: Changed this to test out the lint rule updates and stuff. 🤷🏻‍♂️

src/gateway/__tests__/request.test.js
```diff
@@ -27,7 +27,7 @@ describe("#abortInFlightRequests", () => {
             finally: jest.fn().mockReturnThis(),
         };
         jest.spyOn(MakeRequest, "makeRequest").mockReturnValue(fakeRequest);
-        request(fakeOptions, fakeLogger, "URL");
+        request(fakeLogger, "URL", fakeOptions);
```
somewhatabstract: They are optional now, so they have to go at the end. This is more idiomatic, I think, anyway.

src/gateway/request.js
```diff
@@ -40,10 +47,15 @@ export const abortInFlightRequests = (): void => {
  * have the abort function. Therefore, you'll need to readd it.
  */
 export const request = (
-    options: RequestOptions,
     logger: Logger,
     url: string,
+    options?: RequestOptions,
```
somewhatabstract: They are optional now, so they have to go at the end. This is more idiomatic, I think, anyway.

src/gateway/run-server.js
```diff
@@ -44,6 +44,8 @@ export const runServer = async (
          */
         .get("/render", asyncHandler(makeRenderHandler(renderFn)));
 
+    app.set("trust proxy", true);
```
somewhatabstract: Was reading up on GAE stuff and this was one of the things I added to try and get HTTPS things working.

https://cloud.google.com/appengine/docs/standard/nodejs/runtime#https_and_forwarding_proxies

We probably don't need it anymore, but I also don't think it's a problem to keep it.

src/shared/get-agent-for-url.js
```diff
@@ -0,0 +1,60 @@
+// @flow
```
somewhatabstract: This is based on https://github.com/Khan/webapp/blob/4f747e6730e41a8e6025bcf7e94b92f671cf6df0/services/graphql-gateway/src/http.js

src/shared/start-gateway.js
```diff
@@ -62,4 +62,39 @@ export async function startGateway<
...
+        /**
+         * The Flow type for a Node server does not include headersTimeout.
+         * However, if we don't do the following shenanigans, it puts an error
+         * on the appWithMiddleware.listen call above instead of here, and that
+         * just feels wrong. I tried a $FlowIgnore here, but that doesn't work,
+         * it has to be surpressed above in that case.
+         */
+        const trickFlow: any = gateway;
+        trickFlow.headersTimeout = gateway.keepAliveTimeout + 5000;
+    }
```
somewhatabstract: This is based on https://github.com/Khan/webapp/blob/4f747e6730e41a8e6025bcf7e94b92f671cf6df0/services/graphql-gateway/src/main.js but with some changes to support an option and because, I am guessing, the differences in flow versions.

src/gateway/__tests__/request.test.js
jeresig: Yep! Like this, thank you.

src/gateway/run-server.js
jeresig: If we're keeping this can you add a comment with the link you included above? Thanks!

src/gateway/types.js
```diff
@@ -280,6 +197,15 @@ export type RenderGatewayOptions = {
      */
     +port: number,
 
+    /**
+     * Optional value in milliseconds for keepalive timeout of the server.
+     * For running in Google Cloud, this should be higher than the load
+     * balancer's own keepalive timeout value, which at time of writing was
+     * 80000ms.
```
jeresig: Might be good to include a link to where that timeout is documented in case we ever need to update it!

src/shared/types.js
```diff
@@ -141,6 +141,15 @@ export type GatewayOptions = {
      * What runtime mode the gateway is running under.
      */
     +mode: Runtime,
+
+    /**
+     * Optional value in milliseconds for keepalive timeout of the server.
+     * For running in Google Cloud, this should be higher than the load
+     * balancer's own keepalive timeout value, which at time of writing was
+     * 80000ms.
```
jeresig: (Same note here as above, adding a link to where this is documented)

PR #88, opened 2020-03-31
Begin implementation of JSDOM render environment

somewhatabstract:
# Summary

This is the start of the "render environment" concept. For now, we have one environment, but the aim is to make it reasonably easy to support others by having a common interface.

So, I've added the new `JSDOMSixteenEnvironment` with everything upto downloading entrypoint files in place. This means it does everything the latest webapp use of `render-gateway` does - there is [a corresponding diff for webapp](https://phabricator.khanacademy.org/D61967) to show how this now replaces that implementation.

Currently, the environment doesn't do anything else. That'll be the next diff; I just felt this was a good stopping point for a review.

## Issues

WEB-2044

## Review Guidelines

I'll add some inline comments where I think it's appropriate or helpful to guide the review.


flow-typed/custom/jsdom.js
```diff
@@ -0,0 +1,332 @@
+// flow-typed signature: 6393943152251b303dd7a2ad2a2fff94
```
somewhatabstract: This file is copied from the version I had generated in `react-render-server`.

flow-typed/custom/jsdom.js
```diff
@@ -0,0 +1,332 @@
...
+
+declare interface jsdom$ResourceLoaderConstructorOptions {
+    strictSSL?: boolean;
+    proxy?: string;
+    userAgent?: string;
+}
+
+declare class jsdom$ResourceLoader {
+    constructor(obj?: jsdom$ResourceLoaderConstructorOptions): this;
+    fetch(url: string, options?: jsdom$FetchOptions): ?Promise<Buffer>;
```
somewhatabstract: Unlike the `react-render-server` version, this makes `options` optional. I checked the actual JSDOM code, and this is the case.

package.json
```diff
@@ -82,5 +82,8 @@
   "resolutions": {
     "**/**/braces": "^2.3.1",
     "**/**/minimist": "^1.2.1"
+  },
+  "peerDependencies": {
+    "jsdom": "^16.2.2"
```
somewhatabstract: I think it makes sense for this to be a peer rather than a full dependency.

run_local.js
```diff
@@ -25,6 +23,13 @@ async function main() {
                 status: 200,
                 headers: {},
             }),
+    };
+
+    runServer({
+        name: "DEV_LOCAL",
+        port: 8080,
+
+        renderEnvironment,
```
somewhatabstract: This hopefully illustrates a little how this shifts the architecture. Instead of passing a single function for rendering and having the gateway manage everything around that, we now pass an environment that is responsible for that management, providing us a neat abstraction.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-environment.js
```diff
@@ -0,0 +1,127 @@
...
+         *    - call the afterEnvSetup and attach anything it returns to the
+         *      vm context
+         *
+         * 2. Need to setup the environment with render registration callbacks
+         *    - see the render function in render.js of RRS
+         *
+         * 3. Invoke the render and return its result.
+         *
+         * 4. Finally, close the JSDOM environment.
+         */
```
somewhatabstract: A summary of the things that are yet to occur.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-environment.js
```diff
@@ -0,0 +1,127 @@
...
+            const files = await this._configuration.getFileList(url, renderAPI);
+
+            /**
+             * Now let's use the resource loader to get the files.
+             * We ignore the `FetchOptions` param of resourceLoader.fetch as we
+             * have nothing to pass there.
+             */
+            return await Promise.all(
+                files.map((f) => {
+                    const fetchResult = resourceLoader.fetch(f);
```
somewhatabstract: Unlike with the `react-render-server`, I realised that we can reuse the custom resource loader for our file downloads here, simplifying our environment configuration. It also makes this more versatile, as the custom resource loader could just be loading files from disk for all this code cares.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -0,0 +1,144 @@
...
+import {applyAbortablePromisesPatch} from "./apply-abortable-promises-patch.js";
+
+/**
+ * A ResourceLoader implementation for JSDOM sixteen-compatible versions of
+ * JSDOM that only allows for fetching JS files.
+ *
+ * A JS file request is identified by the regular expression:
+ *   /^.*\.js(?:\?.*)?/g
+ */
+export class JSDOMSixteenResourceLoader extends ResourceLoader {
```
somewhatabstract: This is pretty much identical to the `CustomResourceLoader` of `react-render-server`, except for a couple of tweaks.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -0,0 +1,144 @@
...
+     * Used to indicate if any pending requests are still needed so that we
+     * can report when an unused request is fulfilled.
+     */
+    _active: boolean;
+    _renderAPI: RenderAPI;
+    _requestOptions: RequestOptions;
+
+    static get EMPTY_RESPONSE(): Promise<Buffer> {
+        return Promise.resolve(Buffer.from(""));
+    }
```
somewhatabstract: This is now an accessor so that it creates a new value on call. That way, it will definitely get an abortable promise.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @param {RequestFn} requestFn
+     * The function responsibly for fulfilling GET requests for URLs.
+     */
+    constructor(
+        renderAPI: RenderAPI,
+        requestOptions?: RequestOptions = DefaultRequestOptions,
+    ) {
+        // Patch before super to make sure promises get an abort.
+        applyAbortablePromisesPatch();
```
somewhatabstract: We make sure that the patch is applied here.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -0,0 +1,144 @@
...
+             * or resolve of a promise, for example.
+             *
+             * If it's inlinedata, it really doesn't matter, so let's log it
+             * only if it's for a file.
+             */
+            if (!isInlineData) {
+                logger.warn(
+                    `File fetch attempted after resource loader close: ${readableURLForLogging}`,
+                );
+            }
```
somewhatabstract: Unlike the `react-render-server`, we only output this warning for real files. Seeing one for inline data was uninformative and really of no consequence.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -0,0 +1,144 @@
...
+             */
+            return JSDOMSixteenResourceLoader.EMPTY_RESPONSE;
+        }
+
+        /**
+         * This must be a JavaScript file request. Let's make a request for the
+         * file and then handle it coming back.
+         */
+        const abortableFetch = request(logger, url, {
+            ...this._requestOptions,
```
somewhatabstract: Because the request options configure caching and such, we can guarantee that we share superagent things with any other request calls that use the same options.

src/gateway/handlers/__tests__/make-render-handler.test.js
```diff
@@ -155,15 +167,111 @@ describe("#makeRenderHandler", () => {
...
+                {
+                    logger: fakeLogger,
                     getHeader: expect.any(Function),
                     trace: expect.any(Function),
-                }),
+                },
             );
         });
 
+        describe("provided render API", () => {
```
somewhatabstract: Filled in some testing gaps with tests that will be useful even after we implement proper header tracking (also fills in that 0.3% deficit in our coverage).

src/gateway/types.js
```diff
@@ -91,14 +92,16 @@ export interface AbortablePromise<T> extends Promise<T> {
...
 
-/**
- * Callback to begin a trace session.
- *
- * @param {string} name The name of the traced action.
- * @returns {ITraceSession} A trace session that the caller should use to
- * indicate when the session is finished.
- */
-export type TraceCallback = (name: string) => ITraceSession;
+export interface TraceCallback {
```
somewhatabstract: I was toying with getting better editor integration for function calls not typed inline at their usage points.

src/gateway/types.js
```diff
@@ -258,7 +236,7 @@ export type RequestOptions = {
     /**
      * The agent to be used for the request.
      */
-    +agent?: Agent,
+    +agent?: HttpAgent | HttpsAgent,
```
somewhatabstract: Now we're using `getAgentFromURL`, this needs to support either.

src/gateway/environments/jsdom-sixteen/index.js
```diff
@@ -0,0 +1,12 @@
+// @flow
```
somewhatabstract: I put this together because I envisage that long term, environments would probably be their own packages so that users of our `render-gateway` implementation can install only the one they want to use. I feel this keeps things clearly grouped in a manner that makes sense and also makes it easier to namespace this stuff in the final API we currently expose from the root gateway index

package.json
jeresig: Hmm, potentially! I think that for now having it be a proper dependency is fine - and in the future if we have multiple render environment "plugins" then each plugin can have its own dependencies (specific versions of jsdom, etc.)

src/gateway/environments/jsdom-sixteen/index.js
jeresig: Yeah, I like this! It'll make future JSDOM upgrades, or other environment experiments, a lot easier to coordinate, as well. Musing: I wonder if we might want a way to configure with env is used at request time (so that we could A/B test different environments, potentially).

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
jeresig: Do we think that this is something we'll need to do for every environment we have? Would it be worth abstracting this out?

src/gateway/environments/jsdom-sixteen/index.js
somewhatabstract: Yeah, there's nothing preventing a gateway implementation doing that with environments if it so chooses. It could, for example, create an environment that switches environments based on render.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
somewhatabstract: I don't think we will need to, no. This is specifically a hack for JSDOM and one that we soon won't need as they just fixed the underlying issue about 2 days ago.

package.json
somewhatabstract: I did this mostly so that we don't have to come here and update this repo, then update webapp just to make a simple JSDOM dep update for this stuff.

PR #101, opened 2020-04-09
Implement JSDOM rendering

somewhatabstract:
## Summary

Well, this is exciting. With this change we now have a fully operational `render-gateway`, in so much that it could do what the current one does. Granted, we still need to handle headers and things, but this is still exciting.

I wrote the code for this diff about 2.5 times and in the process learned ways I had overcomplicated things in the RRS and also how not to be too clever. 😄 

In this diff we implement the JSDOM environment fully to load JS and execute the render. The bulk of the work here was getting tests written. I would have liked to split the environment up a bit more into testable pieces, but I couldn't come up with a division that really worked; the code is easier to follow as it is, I feel, so the tests just had to be a little more complex. I wrote them as a story from the top of the render method to the bottom, which helps. Thank goodness I wrote them though; they found at least 2 bugs.

## Issues

WEB-2044


flow-typed/custom/jsdom.js
```diff
@@ -34,11 +34,11 @@ declare export class jsdom$JSDOM {
...
-     * Behind the scenes, a jsdom Window is indeed a VM context.
-     * To get access to this ability, use the runVMScript() method.
+     * which can be compiled ahead of time and then run multiple times on a
+     * given "VM context".
+     * To get access to this ability, use the getInternalVMContext() method.
      */
-    runVMScript(script: vm$Script): any;
+    getInternalVMContext(): vm$Context;
     reconfigure(settings: jsdom$ReconfigureSettings): void;
 }
```
somewhatabstract, line 43: JSDOM 16 changes how the VM is accessed. I did not identify any other type changes, so I just tweaked the typelib for this.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-configuration.js
```diff
@@ -6,6 +6,7 @@ import type {IJSDOMSixteenConfiguration} from "./types.js";
  * environment.
  */
 export class JSDOMSixteenConfiguration implements IJSDOMSixteenConfiguration {
+    +registrationCallbackName: string;
```
somewhatabstract, line 9: Rather than having a magic string, I thought it better if the webapp implementation of the `render-gateway` service could specify what the registration callback should be. That way we can rework that all inside the webapp repo.

src/gateway/environments/jsdom-sixteen/patch-against-dangling-timers.js
```diff
@@ -0,0 +1,67 @@
...
+import type {IGate, ITimerAPI} from "./types.js";
+
+/**
+ * Patch the timer API to protect against dangling timers.
+ *
+ * @returns {IGate} A gate API to control when timers should be allowed to run
+ * (gate is open), or when we should prevent them running and report dangling
+ * timers (gate is closed).
+ */
+export const patchAgainstDanglingTimers = (objToPatch: ITimerAPI): IGate => {
```
somewhatabstract, line 11: This is mostly copied from RRS except now, rather than a flag hanging off the VM context, we use a gate interface to turn things on and off.

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-configuration.js
jeresig: Oh nice, I like that!

src/gateway/handlers/make-render-handler.js
```diff
@@ -36,8 +36,8 @@ async function renderHandler(
     };
 
     /**
-     * TODO(somewhatabstract, WEB-2057): Hook in tracing (make sure that we
-     * don't leave trace sessions open on rejection (or otherwise)).
+     * TODO(somewhatabstract, WEB-2057): Mmake sure that we don't leave trace
```
jeresig, line 39: *Make

src/gateway/environments/jsdom-sixteen/__tests__/jsdom-sixteen-environment.test.js
```diff
@@ -247,18 +279,739 @@ describe("JSDOMSixteenEnvironment", () => {
...
+            const environment = new JSDOMSixteenEnvironment(fakeConfiguration);
+
+            // Act
+            const result = await environment.render("URL", fakeRenderAPI);
+
+            // Assert
+            expect(result).toEqual({
+                body: "THIS IS A RENDER!",
+                status: 200,
+                headers: {},
```
jeresig, line 835: Woohoo!

PR #110, opened 2020-04-20
[WEB-1870] Include headers in response

somewhatabstract:
## Summary

This takes the headers from the render result and puts them on the response. It also validates 3xx headers to ensure that those that require it have a Location header.

## Issues

This resolves WEB-1870.

Additional webapp work needed to fully integrate this work: that will be under WEB-1108.


somewhatabstract:
> Perfect! So I'll use `getTrackedHeaders` and then use it to build up the correct headers that we return. Looks good to me!

You won't need to in that `KAServer` will be where you access this; and I'll just map it to the existing getUsedHeaders call it exposes. So as far as you're concerned, this should be transparent.

jeresig:
@somewhatabstract Ahh, ok! Sounds good, thank you.

PR #117, opened 2020-04-24
Support canvas hack and fix abort issue

somewhatabstract:
## Summary

Yay! The `render-gateway` works in a production environment. However, two bugs became quickly apparent. One we planned for (WEB-1393) and one unexpected (WEB-2144). This fixes both.

Added test cases for the WEB-2144 scenario too, which I ran with the old code and the new to check that it fails/passes accordingly.

## Issues

WEB-1393
WEB-2144


somewhatabstract:
See D62594 for the webapp side to this

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "render-gateway",
-  "version": "0.0.3",
+  "version": "1.0.0",
```
jeresig, line 3: Yes! :)

PR #129, opened 2020-05-04
[WEB 2085] Add custom render error handling

somewhatabstract:
## Summary

Add a configurable handler to custom handle render errors.

## Issues

WEB-2085


src/gateway/handlers/handle-error.js
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * The calling code should handle if the original request was valid or
+     * not, so we just appease flow here.
+     */
+    const requestURL = typeof req.query.url === "string" ? req.query.url : "";
+
+    /**
+     * Before we return the basic 500 error, let's give our configuration
+     * chance to make a nicer error page.
```
jeresig, line 46: *a chance

src/gateway/handlers/handle-error.js
```diff
@@ -0,0 +1,89 @@
...
+        );
+        if (overriddenResponse != null) {
+            logger.error(`${overallProblem}; custom error response generated`, {
+                ...simplifiedError,
+                requestURL,
+            });
+            res.send(overriddenResponse);
+            return;
+        }
+    } catch (e2) {
```
jeresig, line 62: e2? Might be nice to come up with a clearer variable name

src/gateway/handlers/handle-error.js
```diff
@@ -0,0 +1,89 @@
...
+         * Oh no, our configuration threw too!
+         * Ouch. We should report this.
+         */
+        const innerError = extractError(e2);
+        logger.error(`${overallProblem}; custom handler failed`, {
+            ...innerError,
+            originalError: simplifiedError,
+            requestURL,
+        });
+        res.json({
```
jeresig, line 73: Huh, why are we returning `json` here and below? It seems like we'd want a simple HTML page, at least.

src/gateway/handlers/handle-error.js
```diff
@@ -0,0 +1,89 @@
...
+     * not, so we just appease flow here.
+     */
+    const requestURL = typeof req.query.url === "string" ? req.query.url : "";
+
+    /**
+     * Before we return the basic 500 error, let's give our configuration
+     * chance to make a nicer error page.
+     */
+    try {
+        const overriddenResponse = errorHandler?.(
```
jeresig, line 49: How do we expect that this function to set response headers - by modifying the headers object? It seems like we should set the `Content-type: text/html` header by default, at least, so that the handler doesn't have to do it. But I do think that the error response handler is going to want to have the same functionality as the main render function - we want to be able to compute the vary headers and set the right Surrogate-Control headers so that Fastly caches the response correctly, otherwise users will just keep making failed requests again-and-again.

somewhatabstract: Do we set headers on 500 errors like that?

jeresig: Yes, for errors handled internally we'll set those headers (same as we would for 404, 403, or other error pages) - if we have the opportunity to generate an error page it feels like we'd want to try and do it there, too!

somewhatabstract: OK, makes sense.

src/gateway/handlers/handle-error.js
somewhatabstract: That's what the code was already doing for unhandled errors, so I didn't see it as relevant to this change to do anything specific about that.

Seems like a different code change, including decisions on how/where we specify that HTML (is it here, in the base implementation, or a configurable field? How do we cope with a bad configuration? etc.)

src/gateway/handlers/handle-error.js
somewhatabstract: So, as I understand it, we have three main ways to error:

1. Render causes error, but webapp code catches and provides a nice 500 result with headers; this code never even knows; it just sends back the render result.
2. Render fails and we need to provide something half decent so we call the custom handler and the render-gateway service builds a nice result from that specific to webapp - Does this add minimal headers to ensure result is cached, assuming that it's going to happen every time for the same request? - what if it's a one off? Do we want to cache the result? I'm not sure - this bit us with those 502 errors as we cached nothing and then never SSR'd when it was only a temporal issue; that said, seems reasonable we still might want to.
- I ask because this is not going to use webapp frontend code; it will all be in the service, so it's not intended to be super complex. 
3. The custom handling fails and we just have to return something. At this point, we're likely not adding any headers since that's true last ditch and this doesn't know anything about specifics.

src/gateway/handlers/handle-error.js
somewhatabstract: I'll tackle this in a different diff. I've added a TODO

PR #133, opened 2020-05-08
[WEB-2113] Remove renderAPI hack in favour of afterEnvSetup

somewhatabstract:
## Summary

This gets rid of our hacky way to pass the renderAPI calls into the render callback. Instead, we leave it up to the environment configuration to work that out with its afterEnvSetup and take the implementation's opinions out of the equation.

## Issues

WEB-2113


src/gateway/environments/jsdom-sixteen/jsdom-sixteen-environment.js
```diff
@@ -239,23 +236,11 @@ export class JSDOMSixteenEnvironment implements IRenderEnvironment {
...
-            /**
-             * And now we run the registered callback inside the VM, passing
-             * the render API pieces we just provided.
+             * And now we run the registered callback inside the VM.
              */
             const result: RenderResult = await runScript(`
     const cb = window["${registrationCallbackName}"]["${registeredCbName}"];
-    const {trace, getHeader} = window["${registrationCallbackName}"]["${renderAPIName}"];
-    cb(getHeader, trace);`);
+    cb();`);
```
jeresig, line 243: This could be simplified a bit now that we no longer have to transfer the variables/arguments - could be: `window["${registrationCallbackName}"]["${registeredCbName}"]();`

PR #136, opened 2020-05-13
WEB-2281: Rename /render to /_render

jeresig:
This will allow us to have a different URL from react-render-server so that we can more easily gate traffic to it in dispatch.yaml.

I'm going to have a separate diff where I integrate this into webapp.

Issue: https://khanacademy.atlassian.net/browse/WEB-2281

Test plan:
I ran `yarn start`

And then inside webapp I generated the headers file and ran:

`curl http://localhost:8080/_render?url=http://localhost:8081/ -H @headers.json`

And I got back:

```
You asked us to render http://localhost:8081/
```


somewhatabstract:
LGTM

PR #139, opened 2020-05-21
WEB-2289: Vary on headers even if they don't exist

jeresig:
I figured out the issue we were hitting! We weren't tracking headers that we accessed that didn't exist. It turns out it doesn't matter if they exist, or not, we still need to Vary on them. This is part of the Vary specification (you can Vary on headers that don't exist). More info:
tools.ietf.org/html/rfc7231#section-7.1.4

```
   For example, a response that contains

     Vary: accept-encoding, accept-language

   indicates that the origin server might have used the request's
   Accept-Encoding and Accept-Language fields (or lack thereof) as
   determining factors while choosing the content for this response.
```

This was causing issues as the various KAAC-derived headers aren't set at all when there is no KAAC cookie, so them being missing meant that we never tracked them and thus never Varied on them - which caused us to cache the logged-out version of a page and never attempt to render a new version when the user was logged in.

Issue: https://khanacademy.atlassian.net/browse/WEB-2289

Test plan:
I ran the tests and they passed. The Vary logic is handled in Webapp but I believe that this will work fine based on this change.


somewhatabstract:
Oh of course! Doh! Great fix. Thanks for working this out.

PR #157, opened 2020-06-01
[WEB-2306] Log request info like headers

somewhatabstract:
## Summary

This adds a middleware that logs some info about the request being handled. I did this as middleware, that way we get useful log entries if say a bad request comes in or we typo something, or whatever.

## Issues

WEB-2306


src/gateway/run-server.js
```diff
@@ -44,6 +45,12 @@ export const runServer = async (
          * a known source.
          */
         .use(await makeCheckSecretMiddleware(authentication))
+        /**
+         * After the secret check, we log info about the request. Since this
+         * is logging, it MUST go after the secret check or we might leak a
+         * secret, and we don't want that.
+         */
+        .use(logRequestInfoMiddleware)
```
kevinbarabash, line 53: Does the secret check delete the secret?

src/gateway/middleware/__tests__/log-request-info-middleware.test.js
```diff
@@ -0,0 +1,65 @@
...
+        const fakeLogger = {
+            debug: jest.fn(),
+        };
+        const fakeRequest: any = {
+            url: "URL",
+            headers: {
+                HEADER: "VALUE",
+            },
+            method: "GET",
+        };
```
kevinbarabash, line 55: This could be an empty object since we don't care what gets logged for this test.

PR #160, opened 2020-06-03
Fix secret redaction

somewhatabstract:
## Summary

Turns out the code that was supposed to delete the secret header wasn't. I spotted this during dev testing of other changes. The issue is that the `req.header()` function is case-insensitive, but the `req.headers` object uses lowercase keys. We were not normalizing the header name before trying to delete it, and worse, we weren't validating that the delete succeeded.

This adds those things in and tests for them to ensure we don't regress.


somewhatabstract:
Yes, it will

somewhatabstract:
Coverage dropped because it thinks line 37 isn't covered, but there's a test that explicitly captures that error, so it definitely is.

somewhatabstract:
Just realised some test cases were missing or needed updating. Just adding those

PR #163, opened 2020-06-04
[WEB-2359] Ensure agents free sockets between requests

somewhatabstract:
## Summary

This change ensures that our request agents are destroyed after each request. This ensures their sockets are closed. Without this, it seems that they retain references that lock our downloaded JS into memory for a long time. While they may ultimately get released, it's usually too late and we run out of memory.

With this change, after each JSDOM environment is torn down, any created agents are destroyed. This appears to fix the bulk of retention issues (though it's hard to be certain since the dev tools also cause a global retention in the file dumps).

## Issues

WEB-2359


src/shared/get-agent-for-url.js
```diff
@@ -25,34 +25,25 @@ import type {Agent as HttpsAgent} from "https";
...
- * Therefore, the returned agent (one per protocol - http or https) will be
- * the same on repeated requests.
+ * Get a new agent to use for a given URL.
  *
- * Agents are created on first request.
+ * For keep-alive behavior to work, the agent must be shared across requests
+ * but this is left for the calling code to manage. We do this because once the
+ * agent isn't needed (such as at the end of a request), we want to destroy it.
+ * Otherwise, it can hang around keeping sockets open which can lead to
+ * retaining large chunks of memory and creating memory leaks.
```
jeresig, line 35: Very interesting!

PR #178, opened 2020-06-15
Handle gracefully shutting down the process, if possible

jeresig:
This will make it easier for us to use pm2 to manage the server, if we want.

Issue: https://khanacademy.atlassian.net/browse/WEB-2469


src/shared/start-gateway.js
```diff
@@ -63,6 +63,36 @@ export async function startGateway<
...
+                logger.error(
+                    `Error shutting down server: ${
+                        (err && err.message) || "Unknown Error"
+                    }`,
+                );
+                process.exit(1);
+            } else {
+                process.exit(0);
+            }
+        });
```
somewhatabstract, line 93: Should there be a try/catch around this too, just to make sure we catch all possible issues?

src/shared/__tests__/start-gateway.test.js
```diff
@@ -275,4 +275,204 @@ describe("#start-gateway", () => {
             );
         });
     });
+
+    describe("close on SIGINT", () => {
+        it("should do nothing is the gateway doesn't exist", async () => {
```
somewhatabstract, line 280: NIT: `if`

src/shared/start-gateway.js
```diff
@@ -83,6 +83,45 @@ export async function startGateway<
...
+
+        try {
+            gateway.close((err) => {
+                if (err) {
+                    logger.error(
+                        `Error shutting down server: ${
+                            (err && err.message) || "Unknown Error"
+                        }`,
+                    );
+                    process.exit(1);
```
somewhatabstract, line 110: Could you move the `process.exit(1)` to the end of the method and then you wouldn't need it in two places? It would just be the default.

src/shared/start-gateway.js
```diff
@@ -83,6 +83,45 @@ export async function startGateway<
...
+                    process.exit(0);
+                }
+            });
+        } catch (err) {
+            logger.error(
+                `Error closing gateway: ${
+                    (err && err.message) || "Unknown Error"
+                }`,
+            );
+            process.exit(1);
```
somewhatabstract, line 121: See comment above.

src/shared/start-gateway.js
jeresig: I don't think that would work unfortunately - since one is in a callback and one happens synchronously...

jeresig: (to be clear, I tried to make it work but I just ended up with the code duplication anyway)

somewhatabstract: Ah yeah, you're right. I missed that bit. Doh! Thanks for taking a look :)

PR #181, opened 2020-06-16
Allow server to gracefully shutdown if it goes over a memory limit

jeresig:
This makes it so that we can make processes shutdown if they're using too much memory, as specified in an ENV variable.

Issue: none

Test plan:
The tests passed!


src/shared/start-gateway.js
```diff
@@ -120,6 +114,48 @@ export async function startGateway<
...
+                    `Memory usage has gone over maximum. ` +
+                        `(used: ${totalMemory}), limit: ${maxMemory}`,
+                );
+                shutdown();
+            }
+        });
+    }
+
+    /**
+     * In this server is being run using a process manager, such as PM2, we
```
somewhatabstract, line 148: NIT: `In` -> `If`

src/shared/start-gateway.js
```diff
@@ -120,6 +114,48 @@ export async function startGateway<
...
+     * SIGINT signal and then close the server. This prevents new connections
+     * from coming in and waits until the existing connections complete before
+     * the callback is fired. At which point we can safely shutdown the server.
+     * If we fail to respond then the process manager may try to forcefully
+     * shutdown the server after a timeout.
+     */
+    process.on("SIGINT", () => {
+        logger.info("SIGINT received.");
+        shutdown();
     });
```
somewhatabstract, line 159: This feels like it should be in its own function, like `setupShutdownHandling` or something. What do you think?

src/shared/__tests__/start-gateway.test.js
```diff
@@ -276,7 +276,157 @@ describe("#start-gateway", () => {
...
+        const MIN_FREE_MB = process.env.MIN_FREE_MB;
+
+        beforeEach(() => {
+            delete process.env.GAE_MEMORY_MB;
+            delete process.env.MIN_FREE_MB;
+        });
+
+        afterEach(() => {
+            process.env.GAE_MEMORY_MB = GAE_MEMORY_MB;
+            process.env.MIN_FREE_MB = MIN_FREE_MB;
```
somewhatabstract, line 427: This won't work right if the envvars were absent. Instead, it needs to be something like:

```
if (GAE_MEMORY_MB == null) {
    delete process.env.GAE_MEMORY_MB;
} else {
    process.env.GAE_MEMORY_MB = GAE_MEMORY_MB;
}
```

I wrote a helper for this somewhere, but that might've been in some webapp tests.

src/shared/start-gateway.js
```diff
@@ -120,6 +114,48 @@ export async function startGateway<
...
+            // higher than what's allowed and, if so, we shut it down gracefully
+            if (totalMemory >= maxMemory) {
+                logger.info(
+                    `Memory usage has gone over maximum. ` +
+                        `(used: ${totalMemory}), limit: ${maxMemory}`,
+                );
+                shutdown();
+            }
+        });
+    }
```
somewhatabstract, line 145: I like this as middleware. Could you move it to the `middleware` folder. Something like `makeMemoryMonitoringMiddleware` or something?

PR #185, opened 2020-06-18
Implement memory management middleware and shutdown

somewhatabstract:
## Summary

This adds the memory management middleware that instigates shutdown of the gateway if memory limits are exceeded after handling a request.

It also updates jest flow types, moves shared middleware to a consistent location, and has some other smaller tidying.


flow-typed/npm/jest_v26.x.x.js
```diff
@@ -1,5 +1,5 @@
-// flow-typed signature: b074a60bb87f6f6d496e019aec4703b2
-// flow-typed version: 7223a8293e/jest_v25.x.x/flow_>=v0.104.x
+// flow-typed signature: 9a1f9054d272cf6383233b8bfb639f84
+// flow-typed version: 4efeddffd8/jest_v26.x.x/flow_>=v0.104.x
```
somewhatabstract, line 2: Updated via `flow-typed`

src/shared/__tests__/start-gateway.test.js
```diff
@@ -276,203 +288,43 @@ describe("#start-gateway", () => {
         });
     });
 
-    describe("close on SIGINT", () => {
```
somewhatabstract, line 279: The shutdown stuff is encapsulated in a separate call now.

src/shared/shutdown.js
```diff
@@ -0,0 +1,48 @@
...
+                );
+            } else {
+                logger.info("Gateway closed. Shutting down process.", () =>
+                    resolve(),
+                );
+            }
+        });
+    })
+        .then(() => process.exit(0))
+        .catch(() => process.exit(1));
```
somewhatabstract, line 48: Put these here as it seemed neater and also means that the catch will cover things beyond just where we explicitly call reject.

src/shared/start-gateway.js
```diff
@@ -144,6 +123,8 @@ export async function startGateway<
      * [3] https://khanacademy.slack.com/archives/CJSE4TMQX/p1573252787333500
      */
     if (gateway != null) {
+        gatewayStarted(gateway);
```
somewhatabstract, line 126: Makes sure we have a gateway to shutdown when `shutdownGateway` is called.

flow-typed/npm/winston_v3.x.x.js
```diff
@@ -81,7 +81,9 @@ declare type $winstonLoggerConfig<T: $winstonLevels> = {
 };
 
 declare type $winstonLogger<T: $winstonLevels> = {
-    [$Keys<T>]: (message: string, meta?: Object) => void,
+    // Start: https://github.com/Khan/render-gateway/pull/??
+    [$Keys<T>]: (message: string, metaOrCallback?: Object | () => void, callback?: () => void) => void;
+    // End: https://github.com/Khan/render-gateway/pull/??
```
somewhatabstract, line 86: I will update the comments with this PR number, now it has one.

This means we properly support the callback method that ensures all transports have received a logged message.

src/shared/use-app-engine-middleware.js
```diff
@@ -19,11 +21,13 @@ export async function useAppEngineMiddleware<TReq: $Request, TRes: $Response>(
...
             .use(await makeRequestMiddleware(mode, logger))
-            // Add requestID middleware
+            // Add requestID middleware.
             .use(makeAppEngineRequestIDMiddleware(logger))
             // Add the app.
             .use(app)
             // Add the error logging middleware.
             .use(makeErrorMiddleware(logger))
+            // Add memory monitoring.
+            .use(asyncHandler(makeMemoryMonitoringMiddleware(logger)))
```
somewhatabstract, line 31: Need to modify this bit to just not add it if the env vars aren't there at setup time. I'll make that change.

src/shared/shutdown.js
```diff
@@ -31,6 +31,13 @@ export const shutdownGateway: (logger: Logger) => Promise<void> = (logger) =>
             return;
         }
         logger.debug("Closing gateway.");
+
+        // We notify the process manager that we're about to go offline
+        // so that it stops sending us new requests.
+        if (process.send) {
```
somewhatabstract, line 37: Huh, is this likely to be undefined?

src/shared/shutdown.js
```diff
@@ -51,6 +58,15 @@ export const shutdownGateway: (logger: Logger) => Promise<void> = (logger) =>
...
         });
+
+        // If the server hasn't shutdown in 15s then we force it to shutdown
+        // This could be because of connections being kept alive
+        setTimeout(() => {
+            logger.error(
+                "Server failed to close. Forcing it to shutdown.",
+                () => reject(),
+            );
+        }, 15000);
```
somewhatabstract, line 69: Should this timeout be configurable so that we can easily tweak it in settings from the webapp implementation?

src/shared/shutdown.js
somewhatabstract: Oh, perhaps when not executed via PM2 or a similar management app.

PR #196, opened 2020-06-23
Add examples and fixes from #185

jeresig:
Brings over things that we still want to land!

Issue: none

Test plan:
Ran `yarn test` and it passed.


flow-typed/npm/winston_v3.x.x.js
```diff
@@ -81,7 +81,13 @@ declare type $winstonLoggerConfig<T: $winstonLevels> = {
...
 
 declare type $winstonLogger<T: $winstonLevels> = {
-    [$Keys<T>]: (message: string, meta?: Object) => void,
+    // Start: https://github.com/Khan/render-gateway/pull/185
+    [$Keys<T>]: (
+        message: string,
+        metaOrCallback?: Object | (() => void),
+        callback?: () => void,
+    ) => void,
+    // End: https://github.com/Khan/render-gateway/pull/185
```
somewhatabstract, line 90: NIT: Probably need to update this to #196

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "start:example": "bash -c 'RUNNER=\"./examples/$0/run.js\"; KA_LOG_LEVEL=silly NODE_ENV=development babel-watch \"$RUNNER\"'",
```
somewhatabstract, line 14: Query: I toyed with this locally, but what do you think about changing `start:example` to just `example`? Since I've been using this a lot, it's been quicker to type :grinning:

jeresig: Sure thing!

somewhatabstract:
Looks like the branch name is confusing the checkout action. Could fix by changing the branch name or editing the action to add a `--`

somewhatabstract:
Raised WEB-2495 to cover the publish action issue

PR #203, opened 2020-06-25
WEB-2510: Pass headers through in render API

jeresig:
## Summary:

We're going to ask that the render environment handle all things header-related (including tracking header access) so that we don't have to try and manage this API across process boundaries.

Issue: WEB-2510

## Test plan:

`yarn test`


src/gateway/types.js
```diff
@@ -147,22 +130,12 @@ export type RenderResult = {
...
-     * should contain. It is provided as a callback so that the gateway
-     * implementation can track which headers influence a render, which can then
-     * be reported back as a Vary header in the gateway response.
-     */
-    +getHeader: GetHeaderCallback,
-
-    /**
-     * Callback to retrieve a map of the request headers that have been accessed
-     * and their values.
+     * Access of these headers should be tracked and a Vary header should be
```
somewhatabstract, line 135: Can you explicitly state that all header access must be tracked and added to Vary, not just those that exist?

jeresig:
@somewhatabstract As far as I can tell all the other changes will be happening inside the code in Webapp, the changes here are pretty minimal!

somewhatabstract:
> @somewhatabstract As far as I can tell all the other changes will be happening inside the code in Webapp, the changes here are pretty minimal!

Awesome.

PR #219, opened 2020-07-01
Add JSDOM15 environment

somewhatabstract:
## Summary:
This adds support for the JSDOM 15 environment.

TODO: Fix tests. Pushed this for now so I can test if it's a viable mitigation. No point in doing more work if it doesn't help.

Issue: WEB-2498

## Test plan:

`yarn test`


somewhatabstract:
@jeresig abandoning this diff since it didn't help us, it seems.

Should be easy enough to recreate if we ever needed to.

PR #224, opened 2020-07-02
Make memory usage traces more useful

somewhatabstract:
## Summary:

This is just a small change to make our memory usage logging during traces
a little easier on the eyes.

Instead of `memoryBefore` and `memoryAfter`, we will now have `/memory/final`
and `/memory/delta`. The first is the `memoryAfter` values renamed, the delta
is the difference of `memoryBefore` and `memoryAfter`.

Issue: WEB-2579

## Test plan:
`yarn test`
`yarn example logging`


src/shared/trace.js
```diff
@@ -81,7 +82,6 @@ export const trace = (
          * We disable this lint rule as the linter does not appear to
          * understand the optional chaining.
          */
-        // eslint-disable-next-line flowtype/no-unused-expressions
```
jeresig, line 84: huh - what fixed this?

somewhatabstract: I think the update to eslint flowtype plugin? It was complaining when I tried to commit until I removed these lines.

PR #225, opened 2020-07-06
Only create one root logger and therefore, only initialize Google Cloud Logging's logging-winston once

somewhatabstract:
## Summary:
This refactors things so that we don't execute `createLogger` on every import
of `getLogger` within a new import context (i.e. a worker).

Issue: WEB-2572

## Test plan:
Test locally with webapp to check general functionality.
Will be putting up a ZND to check in a production environment and verify this fixes the logging issue.


src/gateway/run-server.js
```diff
@@ -75,9 +75,11 @@ export const runServer = async (
     app.set("trust proxy", true);
 
     // Start the gateway.
+    const runtimeMode = getRuntimeMode();
+    const logLevel = getLogLevel();
     const gatewayOptions: GatewayOptions = {
-        mode: getRuntimeMode(),
-        logger: getLogger(),
+        mode: runtimeMode,
+        logger: createLogger(runtimeMode, logLevel),
```
somewhatabstract, line 82: This is part of the fix. We create the logger explicitly at the launch of the gateway.

src/shared/create-logger.js
```diff
@@ -90,7 +90,9 @@ const getTransport = (mode: Runtime): Transport => {
              * to the express request, make sure this transport is passed to
              * that middleware so that it doesn't add its own.
              */
-            return new lw.LoggingWinston();
+            return new lw.LoggingWinston({
+                level: logLevel,
```
somewhatabstract, line 94: Added this just to be sure we're filtering out the right things (though I verified that this was working without the need for this in non-worker-based deploys).

src/shared/get-request-logger.js
```diff
@@ -3,7 +3,14 @@ import type {$Request} from "express";
...
-    defaultLogger: Logger,
+    defaultLogger: ?Logger,
     request?: TReq,
-): Logger =>
-    request != null && request.log != null ? request.log : defaultLogger;
+): Logger => {
+    if (defaultLogger == null) {
+        if (request == null || request.log == null) {
+            throw new Error("No logs available");
+        }
```
somewhatabstract, line 12: Added this since there's now a possibility of there being no root logger.

This should never happen though as the circumstances where that's the case, we only call this with our own instance of a default logger.

src/shared/start-gateway.js
```diff
@@ -29,6 +30,14 @@ export async function startGateway<
...
     }
 
+    /**
+     * Setup logging.
+     * We create the root logger once and then share it via a singleton.
+     * This avoids us creating a new one in each worker, which was happening
+     * when we created the logger on import of `getLogger`.
+     */
+    setRootLogger(logger);
+
```
somewhatabstract, line 40: Instead of creating a logger implicitly in the `getLogger` call, we now set it here based off the options.

src/shared/trace.js
```diff
@@ -1,142 +1,68 @@
 // @flow
-import type {Tracer} from "@google-cloud/trace-agent";
```
somewhatabstract, line 2: Renamed to `trace-impl.js`

src/shared/trace.js
```diff
@@ -1,142 +1,68 @@
...
-import {getGatewayInfo} from "./get-gateway-info.js";
-import {getDelta} from "./get-delta.js";
-import type {Logger, ITraceSession, TraceSessionInfo} from "./types.js";
+import * as traceAgent from "@google-cloud/trace-agent";
```
somewhatabstract, line 2: This was in `KAShared`. Now it's the main export from `Shared`

PR #250, opened 2020-07-31
[WEB-2721] Log url being retried and always buffer

somewhatabstract:
## Summary:
- Knowing what URL is being retried is useful, so now we will
- Always buffer instead of allowing it to be configured. Turns out that we need it and though we might make things more efficient (memory wise) by implementing how that might work for ourselves (such as writing directly to a buffer rather than a string first, then buffer back to string), it's not worth doing now and we will definitely never want to configure buffering off as that just won't work; so let's remove that temptation.

Issue: WEB-2721

## Test plan:
`yarn test`

Updated webapp with this version to try it out and all seems to work well.
We won't know about the retry logging until we see retries, but it is a
simple change, so no concerns there.


src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -174,7 +174,30 @@ export class JSDOMSixteenResourceLoader extends ResourceLoader {
...
+             *
+             * If buffering is on (bad for memory use), then text will always
+             * be present.
+             *
+             * So, let's cater to almost all eventualities.
+             *
+             * There are some circumstances where body might be null too, but
+             * those are cases we should never encounter.
+             */
+            const data = response.text == null ? response.body : response.text;
```
somewhatabstract, line 194: Note that we could use `response.buffered` to determine the buffered state, but this doesn't tell us if any of this is valid. We could also console `response.xhr.response` directly.

somewhatabstract: This is not relevant any more. Resolving.

src/gateway/requests-from-cache.js
```diff
@@ -40,7 +40,7 @@ export const asUncachedRequest = (
      * We need to ensure that what we return has the `abort` method still so
      * that we can let things like JSDOM call abort on promises.
      */
-    const superagentRequest = request.buffer(options.buffer);
+    const superagentRequest = request.buffer(true);
```
somewhatabstract, line 43: We could pass nothing here; that is also equivalent to `true`, but I'd rather be explicit.

src/gateway/requests-from-cache.js
```diff
@@ -105,7 +110,7 @@ export const asCachedRequest = (
             guttedResponse[FROM_CACHE_PROP_NAME] = FRESHLY_PRUNED;
             return guttedResponse;
         })
-        .buffer(buffer);
+        .buffer(true);
```
somewhatabstract, line 113: We could pass nothing here; that is also equivalent to `true`, but I'd rather be explicit.

PR #269, opened 2020-08-26
Make it optional for which google cloud agents we start

somewhatabstract:
## Summary:

Issue: None

## Test plan:

`yarn test`


somewhatabstract:
> Ok! So this is the render-gateway portion of this, I assume there'll be another part for the webapp config of this. Looks good!

Exactly. Though we are on a much newer set of these libraries than the graphql-gateway; perhaps the issue is fixed in the new versions

somewhatabstract:
> perhaps the issue is fixed in the new versions

Doesn't seem like it. Saw comments on https://github.com/googleapis/cloud-debug-nodejs/issues/811 that indicate it's in 5.1.0 too.

PR #270, opened 2020-08-27
Add warmup handler to render-gateway

benjaminjkraft:
We need this to be able to do traffic migration.  It's also probably not
a bad idea in general.

Issue: https://khanacademy.atlassian.net/browse/INFRA-4948

Test plan: yarn test


src/ka-shared/routes/__tests__/make-common-service-router.test.js
```diff
@@ -47,6 +47,48 @@ describe("#makeCommonServiceRouter", () => {
         });
     });
 
+    describe("route /_ah/warmup", () => {
```
somewhatabstract, line 50: Thanks for adding these tests!

somewhatabstract:
Relates to WEB-1990

PR #364, opened 2020-12-18
Make tests dependent on publish

somewhatabstract:
## Summary:

Our actions can take a while. If we have changes to publish, this can mean
extra test runs that are testing the exact same thing. So, let's make tests
dependent on publishes and also make publish be a fail if it actually updates
files, since we know that update will have triggered another run.

## Test plan:
Push this and see what happens. :)


kevinbarabash (approved):
Seems reasonable to me.  Thanks for adding comments in the workflow file.

PR #428, opened 2021-02-24
[FEI-3329].1 Update to Flow 0.142.0

somewhatabstract:
## Summary:

Issue: FEI-3329

## Test plan:
`yarn flow`


src/shared/get-delta.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 type NumberObj = {
     [key: string]: number,
+    ...
```
somewhatabstract, line 4: This didn't fix the issue, so I also suppressed the error for now.

jeresig (approved):

PR #441, opened 2021-03-08
Change request info logging to not write request headers.

latteier:
# Summary

We are currently getting errors from Google due to our log exporting
creating too many bigquery column names. This is causing us to lose
some logs exports. The issue is that the nested headers metadata end
up being converting to one bigquery column per unique column name. Due
to there being so many possible incoming request header names we are
exceeding the max number of columns.

The relevant log exporter is `winston_logs_to_bigquery`
https://console.cloud.google.com/logs/router?project=khan-academy

The solution implemented here is to flatten the headers into a string.
This will avoid writing a new column for each header value.

I've changed the name of the headers column to avoid a type mismatch
issue when changing the existing column type (record) to string.

## Issues

- None

## Review Guidelines

- I'm not sure how to test this manually and would love help on validating that this is safe.


latteier:
> We don't log below info to stack driver at all right now, so moving them to debug will likely mean we don't get them at all. 

Ah we didn't realize that when discussing https://khanacademy.slack.com/archives/CEA6W0F6F/p1615228881136100

> Instead, I propose flattening the headers to a value that doesn't fall foul of this columns issue

That sounds reasonable. Do you have a preferred format for the flattening? If not I'll just pick something. I'm happy to revise the PR.

latteier:
Thanks for the review. I'm going to merge this now.

What's the process to get this into production?

PR #455, opened 2021-03-24
[FEI-2995] Logging enhancements and fixes

somewhatabstract:
## Summary:
Two main changes here.

1. JSDOM console logging is now passed through to the render-gateway logging at the `silly` level. This avoids logging things we don't need in production, but supports them being available for devs as needed.
2. Fixes an issue where we started seeing some double logging  and low level logging in production. Turns out that logging-winston has a fix for a bug we were working around, and this fix meant that our workaround started behaving like the original bug. So, we remove the workaround and now all is well.

Issue: FEI-2995

## Test plan:
`yarn test`

After deployment (staging or otherwise) we should no longer see any `TRACE` messages for downloading files, and we shouldn't see any JSDOM console warn/info/debug/log output (though these are relatively uncommon now).


src/shared/middleware/make-request-middleware.js
```diff
@@ -14,14 +14,11 @@ export const makeRequestMiddleware = <TReq: $Request, TRes: $Response>(
...
-         * log property to the request, allowing us to associate log entries
-         * with a request trace, if the request is being traced.
-         *
-         * WORKAROUND: We must pass the corresponding transport here or it will
-         * get added for us and then we'll have double logging.
+         * In production, we're using the Google logging-winston middleware.
+         * This adds the log property to the request, allowing us to associate
+         * log entries with a request trace, if the request is being traced.
          */
-        return lw.express.makeMiddleware(logger, new lw.LoggingWinston());
```
somewhatabstract, line 24: The change to the `makeMiddleware` function in `src/middleware/express.ts` as part of https://github.com/googleapis/nodejs-logging-winston/pull/553 means that the `LoggingWinston` transport that we add when we create our logger is now properly found and accepted. This means that we don't need to pass one here to trick the middleware into not adding one, and in fact, passing one here now causes issues.

jeresig (approved):
Awesome, thank you for fixing this!

src/gateway/environments/jsdom-sixteen/create-virtual-console.js
```diff
@@ -19,34 +19,38 @@ export const createVirtualConsole = (logger: Logger): VirtualConsole => {
             return;
         }
         const simplifiedError = extractError(e);
-        logger.error(`JSDOM:${simplifiedError.error || ""}`, {
+        logger.error(`JSDOM jsdomError:${simplifiedError.error || ""}`, {
```
kevinbarabash, line 22: I like that we're differentiating JSDOM errors from other errors.

src/gateway/environments/jsdom-sixteen/__tests__/create-virtual-console.test.js
```diff
@@ -69,90 +72,36 @@ describe("#createVirtualConsole", () => {
...
-
-        // Assert
-        expect(fakeLogger.debug).toHaveBeenCalledWith(
-            "JSDOM:This is a debug message",
-            {
-                args: ["and these are args"],
-            },
-        );
-    });
+    it.each(["warn", "info", "log", "debug"])(
```
kevinbarabash, line 82: nice

src/shared/middleware/make-request-middleware.js
kevinbarabash: Do we have to upgrade `@google-cloud/logging-winston` in package.json?

kevinbarabash (approved):
LGTM.  Can the changes to the `lw.express.makeMiddleware` call be tested locally before deploying?

src/shared/middleware/make-request-middleware.js
somewhatabstract: Dependabot already did.

somewhatabstract:
> Can the changes to the lw.express.makeMiddleware call be tested locally before deploying?

It's really hard to test the Google middleware locally.

PR #465, opened 2021-04-05
[FEI-2722] Add new cache tracking approach

somewhatabstract:
## Summary:
This removes the old cache tracking, which only worked for in-memory
cache, and replaces it with something more robust.

A related change will be needed in the webapp usage when this new
version is incorporated since this changes the API.

The way it works now is that the calling code includes a callback in
the request options that returns a cacheID for the request about to be
made. That cacheID is then attached to any fresh response. Therefore, it
can then be used to determine if the response was from the cache or not,
as long as the cacheID is generated in a manner that is unique for a
given batch of requests.

For webapp, we can use the requestID from GAE as our cacheID, therefore
allowing us to see for each request, what was from cache and what was
not, just as we did before.

Issue: FEI-2722

## Test plan:
`yarn test`

Add this to webapp and implement to use requestID as cacheID, then use the
render-gateway and observe that the source label on requests for JS files says
"new request" on the first request, but "cache" for subsequent requests.


jeresig (approved):
Looks ok to me!

PR #473, opened 2021-04-13
[FEI-2726] Aborted requests to give no response

somewhatabstract:
## Summary:
Superagent will return a partial result for an aborted request. If
this then gets executed as JS, we get all sorts of errors that we
really don't care about.

This adds the ability to see the aborted state of a request and use
that to return an empty response instead of whatever partial response
is going to be given.

We use the internal `_aborted` flag of the superagent request because
that seems more reliable than other means available from say the XHR.

Issue: FEI-2726

## Test plan:
`yarn test`
We will need to put this into production traffic where we seemingly face this
problem often so that we can see what impact, if any, it has on the errors that
get logged.


src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -162,10 +162,13 @@ export class JSDOMSixteenResourceLoader extends ResourceLoader {
             agent: this._getAgent(url),
         });
         const handleInactive = abortableFetch.then((response) => {
-            if (!this._active) {
-                logger.silly(
-                    `File requested but never used: ${readableURLForLogging}`,
-                );
+            const aborted = abortableFetch.aborted;
```
jeresig, line 165:
```suggestion
            const {aborted} = abortableFetch;
```

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
```diff
@@ -162,10 +162,13 @@ export class JSDOMSixteenResourceLoader extends ResourceLoader {
...
+            if (!this._active || aborted) {
+                if (!aborted) {
+                    logger.silly(
+                        `File requested but never used: ${readableURLForLogging}`,
```
jeresig, line 169: This would be useful for us to investigate these, I suspect - since these could imply a problem happening somewhere in our code!

jeresig (approved):
Looks good, thank you!

src/gateway/environments/jsdom-sixteen/jsdom-sixteen-resource-loader.js
somewhatabstract: When you say "these", I presume you mean the requests that were made but weren't needed to produce a result, rather than the aborted calls? Or are you suggesting a change to always log, even when requests are aborted?

jeresig: I was interested in the calls that were made but weren't needed.

somewhatabstract: Yeah, looking at the logs may be interesting there.

PR #476, opened 2021-04-15
Update workflow and force new dist

somewhatabstract:
## Summary:
This updates our workflow action versions and makes some code changes to invoke
a new build that will need to push changes to the repo, thus testing the secrets
that I have updated.

Bonus: more test coverage.

Issue: None

## Test plan:
Creating this PR should execute the new workflow


.github/workflows/node-ci.yml
```diff
@@ -48,15 +48,15 @@ jobs:
...
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
     - name: Install dependencies
       run: yarn install --frozen-lockfile
     - name: Run tests with coverage
       run: yarn coverage
     - name: Upload Coverage
-      uses: codecov/codecov-action@v1.0.5
```
somewhatabstract, line 59: Being less specific should mean we'll use the latest version, which in this case, I think we want.

src/shared/ka-error.js
```diff
@@ -12,17 +12,25 @@ export default class KAError<TKind: ErrorKind> extends Error {
...
      * @param {string} message The error message.
-     * @param {TKind} kind The kind of KA error.
+     * @param {TKind} kind The kind of error.
      * @param {Error} [sourceError] The original error that spawned this one.
      */
     constructor(message: string, kind: TKind, sourceError: ?Error = null) {
         super(message);
+
+        // Set the name so we get a nice error output, like
+        // KAInternalError
```
somewhatabstract, line 25: These comments should force the build to generate new source map which will then cause us to push a new dist.

jeresig (approved):
Thanks for these fixes!

PR #477, opened 2021-04-15
[FEI-2726].3 Close everything at once

somewhatabstract:
## Summary:
Before, we were nice and went in reverse order, but this meant that the resource loader was the last thing to close. That meant it's 'I'm no longer serving things' code did not get invoked until after the JSDOM window was closed. Now we close everything 'at once' which should have an impact on some of the errors we see that I suspect are down to the shutdown process.

This also makes the virtual console closeable so that we don't get any further logging noise from the JSDOM environment at close time.

Issue: FEI-2726

## Test plan:
Put this into a ZND and check that it is doing the right thing.  Perhaps load test it there to see if we can induce some errors.

Put it into prod and see what difference it makes.


somewhatabstract:
Put up for diff but I am considering writing some tests to go with this.

jeresig (approved):
This all makes sense to me - hope it's able to work!

somewhatabstract:
@jeresig Flow was very unhappy because of my hackiness. I reimplemented the virtual console as a class and also added some test coverage. I think it likely needs a new review since I changed for files.

.github/workflows/node-ci.yml
```diff
@@ -11,17 +11,25 @@ on:
...
-
+# Our jobs run like this to minimize wasting resource cycles:
+#   1. Prime caches for primary configuration (ubuntu on node 12).
+#      This way the next two jobs can run in parallel but rely on this primed
+#      cache.
+#   2. Lint, coverage, and publish
+#     a. Lint
+#     b. Coverage and publish
+#   3. Run tests for remaining configurations.
+#      Since these don't share caches, we don't need to prime those caches.
```
somewhatabstract, line 22: Overview of the jobs and precedence.

1 -> 2 -> 3

src/gateway/environments/jsdom-sixteen/__tests__/__snapshots__/jsdom-sixteen-environment.test.js.snap
```diff
@@ -1,13 +1,13 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`JSDOMSixteenEnvironment #render should reject if result is malformed 1`] = `"Malformed render result: undefined"`;
+exports[`JSDOMSixteenEnvironment #render should reject if result is malformed ({ body: 'NEED MORE THAN THIS' }) 1`] = `"Malformed render result: {\\"body\\":\\"NEED MORE THAN THIS\\"}"`;
```
somewhatabstract, line 3: I made it so the test case included the data point so it was easier to map from here to the test values. In reality, the snapshots didn't change.

src/gateway/environments/jsdom-sixteen/closeable-virtual-console.js
```diff
@@ -0,0 +1,70 @@
+// @flow
```
somewhatabstract, line 1: Replaces the old `create-virtual-console.js` file.

src/gateway/environments/jsdom-sixteen/closeable-virtual-console.js
```diff
@@ -0,0 +1,70 @@
...
+                method,
+                (message, ...args) =>
+                    !this._closed &&
+                    logger.silly(`JSDOM ${method}:${message}`, {args}),
+            );
+        };
+        passthruLog("warn");
+        passthruLog("info");
+        passthruLog("log");
+        passthruLog("debug");
```
somewhatabstract, line 64: This is identical to what the `create-virtual-console.js` code used to do.

src/gateway/request.js
```diff
@@ -67,16 +67,9 @@ export const request = (
      * to our list of in flight requests.
      */
     const finalizedRequest: AbortablePromise<Response> = (finalizedPromise: any);
-    /**
-     * In tests, we might mock the promise API to return the same mock, so
-     * to avoid cyclic abort calls, we only add abort if we're not the same
-     * object.
-     */
-    if (finalizedRequest !== abortableRequest) {
```
somewhatabstract, line 75: I fixed the tests so that we don't need this hack

.github/workflows/node-ci.yml
```diff
@@ -11,17 +11,25 @@ on:
...
+    name: Prime node_modules cache for primary configuration
     env:
       CI: true
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macOS-latest]
-        node-version: [12.x, 14.x]
+        os: [ubuntu-latest]
+        node-version: [12.x]
```
somewhatabstract, line 32: We only bother priming our primary platform now. This is because the lint job and the publish job run side-by-side and so priming the cache saves them time. However, the test runs that follow the publish and lint jobs are independent of one another and therefore don't benefit from priming at all.

.github/workflows/node-ci.yml
```diff
@@ -72,41 +80,6 @@ jobs:
       run: yarn flow:ci
 
 
-  coverage:
```
somewhatabstract, line 75: This has been rolled into the publish step, which was already running tests anyway.

.github/workflows/node-ci.yml
```diff
@@ -72,41 +80,6 @@ jobs:
...
-    - name: Install dependencies
-      if: steps.node-modules-cache.outputs.cache-hit != 'true'
-      run: yarn install --frozen-lockfile
-    - name: Run tests with coverage
-      run: yarn coverage
-    - name: Upload Coverage
-      uses: codecov/codecov-action@v1
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: true
```
somewhatabstract, line 107: This was the bit that moved to the publish step

.github/workflows/node-ci.yml
```diff
@@ -143,17 +116,22 @@ jobs:
...
     # a non-feature branch.
     if: github.event_name == 'push' || !startsWith(github.head_ref, 'feature/')
     env:
       CI: true
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [12.x]
```
somewhatabstract, line 134: Changed to using a matrix to standardize the yaml and also to make the task names more self-documenting in the Actions summaries.

jeresig (approved):
Looks good, thank you!

PR #480, opened 2021-04-19
Make sure retries are added as a label so they appear in traces AND logs

somewhatabstract:
Issue: FEI-2726

## Test plan:
Putting up a ZND again. It was clear after the last one that I had forgotten
the label.

Running the en loadtest was sufficient to get some examples.
`dev/tools/generate-render-gateway-load-tests.js -c ab -s <ZND> -v <ZND> --with-secrets ./services/render-gateway/__loadtests__/loggedout-en_loadtests.json | parallel '{}'`


src/gateway/request.js
```diff
@@ -51,10 +51,6 @@ export const request = (
      * Then we capture the abort function so we can reattach it later.
      */
     const traceSession = trace(`request`, url, requestLogger);
-    // It would be nice if the metadata of the logger was automatically
-    // added as a trace label, but for child loggers, that's hard since
-    // the child metadata isn't currently accessible off the logger;
-    // the defaultMetadata prop is only the root logger's metadata.
```
somewhatabstract, line 57: I got things the wrong way round about how stuff works - and this comment just wasn't relevant anymore.

src/gateway/request.js
```diff
@@ -74,9 +70,8 @@ export const request = (
             return res;
         })
         .finally(() => {
-            traceSession.end({
-                retries: retryCount,
-            });
+            traceSession.addLabel("retries", retryCount);
```
somewhatabstract, line 73: Changed this from metadata added to the log message into a label that will be included in the log message AND the trace.

PR #484, opened 2021-04-22
Add more data to our error handling

somewhatabstract:
## Summary:

This expands our SimplifiedError type to support any primitive props
the error may expose. This should help us in investigating aborted
requests and our ECONNRESET issues. I dug into them and was unable to
see how we end up with partial JS files still, so hopefully this will
help.

In addition, the log message for late imports is now back to logging at
info as we should track those down in our code and make them smarter
such that they don't get requested if they aren't necessary for the
render (this used to mainly be mathjax but I think we tackled that -
we will soon find out).

Issue: FEI-2726

## Test plan:
`yarn test`

I used Quokka to run this simple test code that forces timeouts to get
an error example.

```
const {request} = require("./src/gateway/request.js");

const {createLogger} = require("./src/shared/create-logger.js");

const logger = createLogger("development", "silly");

const {extractError} = require("./src/shared/extract-error.js");

let tries = 0;
const r = request(logger, "http://example.com", {
    timeout: {
        deadline: 5000,
        response: 30,
    },
    shouldRetry: (...args) => {
        tries++;
        return undefined;
    }
});
//r.abort();
r.aborted //?
r.then(res => {
    tries //?
    r.aborted //?
}).catch(e => {
    e //?
    extractError(e) //?
}).finally(() => console.log("DONE"));
```


jeresig (approved):
Looks good, thank you!

PR #530, opened 2021-06-04
Try adding codeql analysis

somewhatabstract:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master, feature/* ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '33 19 * * 6'
```
kevinbarabash, line 21: How often does this run?  I'm not up on my cron syntax.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
```
kevinbarabash, line 30: neat

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
kevinbarabash, line 35: Does it handle flow?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1
```
kevinbarabash, line 71: Does it analyze the built files?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
```
kevinbarabash, line 68: Remove?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
kevinbarabash, line 57: Does this analyze the project and run `yarn build` automatically?  The comment doesn't mention JavaScript though.

kevinbarabash (approved):
Cool!  I'm curious to see how its output differs from Dependabot.  🚢

.github/workflows/codeql-analysis.yml
somewhatabstract: Ah yeah, I had to look this up. This value was auto-added by the config generator that GitHub provides by default.

The docs are here:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows

Here's a pertinent snippet:
```
┌───────────── minute (0 - 59)
│ ┌───────────── hour (0 - 23)
│ │ ┌───────────── day of the month (1 - 31)
│ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
│ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
│ │ │ │ │                                   
│ │ │ │ │
│ │ │ │ │
* * * * *
```

`33 19 * * 6` means 7:33pm every Saturday.

There are more complex rules than just specifying specific times - more details in the linked docs above.

This shedule trigger exists for this specific workflow because the security flaws it detects might change even if the code doesn't.

.github/workflows/codeql-analysis.yml
somewhatabstract: It does some magic build thing that I didn't quite follow, but it seems to handle it fine.

.github/workflows/codeql-analysis.yml
somewhatabstract: I think so, not entirely sure.

.github/workflows/codeql-analysis.yml
somewhatabstract: I honestly don't know. I think it perhaps only analyzes the source in JavaScript but perhaps knows enough to apply babel transforms. It also supports TypeScript under the javascript setting, so I'm guessing it leverages some sort of build knowledge with source maps.

somewhatabstract: You can see that it found a couple of things already.

https://github.com/Khan/render-gateway/security/code-scanning?query=ref%3Arefs%2Fpull%2F530%2Fmerge+tool%3ACodeQL

PR #734, opened 2022-01-04
[🔥AUDIT🔥] Manually mock cloud profiler

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
A recent update appears to have caused it to start failing due to
the profiler code doing some async action on load :(

Rather than try and hook and await that action, we just manually mock
the whole library.

Issue: XXX-XXXX

## Test plan:
`yarn test`

### DIFF
--- a/src/shared/__tests__/start-gateway.test.js
+++ b/src/shared/__tests__/start-gateway.test.js
@@ -6,6 +6,7 @@ import {createLogger} from "../create-logger.js";
 
 jest.mock("../root-logger.js");
 jest.mock("../setup-stackdriver.js");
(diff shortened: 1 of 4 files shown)


src/shared/__tests__/start-gateway.test.js
```diff
@@ -6,6 +6,7 @@ import {createLogger} from "../create-logger.js";
 
 jest.mock("../root-logger.js");
 jest.mock("../setup-stackdriver.js");
+jest.mock("@google-cloud/profiler");
```
somewhatabstract, line 9: This was unnecessary - but it's also inert.

PR #1065, opened 2023-02-28
Remove debug-agent

somewhatabstract:
## Summary:
This removes the soon to be deprecated Google Cloud debug agent. We don't use it anymore anyway and when we do, it has a massive memory leak.

In addition, I have removed the use of husky and lint-staged. We have generally come to accept that these sort of pre-commit and pre-push gatekeepers can be a nuisance and get in the way of developer flow. Removing them gives us one less thing to maintain - and we have the GitHub Actions that should be checking this stuff.

Issue: FEI-4931

## Test plan:
`yarn flow`
`yarn test`

Not much else to test for this.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "render-gateway",
-  "version": "1.3.0",
+  "version": "2.0.0",
```
somewhatabstract, line 3: **note:** Updated to 2 since this is a breaking change to the types and feature set.

jeresig (approved):

PR #1080, opened 2023-03-28
Update to WS Server

somewhatabstract:
## Summary:
This updates the render-gateway repo to use the new Wonder Stuff Server package, which contains a lot of code that used to be in this repo.

This includes:
- Using Wonder Stuff Code for KindError instead of custom KAError type
- Deleting the code that is no longer needed
- Moving the remaining code to the `src` directory and removing the sub-folders that represented "fake packages"
- Changing the implementation of secrets and request authorization to match the new Wonder Stuff Server implementation
- Updating tests and adding tests
- Renaming the JSDOMSixteen environment to just JSDOM since we have updated through several versions with breaking changes since this was first written; it makes more sense to be just JSDOM based off that experience (this will be its own package in Wonder Stuff post-migration)

Issue: FEI-4960

## Test plan:
`yarn test`
`yarn flow`

I also ran each of the examples to verify that the server worked in general.

Migrated the webapp service to use this version to demonstrate that it is working, and exports all the things it needs to. See https://github.com/Khan/webapp/pull/12812 for the tests steps involved that add confidence to this PR.


jeresig (approved):
OK yeah, this all seems good to me. Thanks for all the updates!

